PR #7, opened 2022-06-17
CONTENT-97 - handle versions display

jlsherrill:


src/components/ContentListTable/ContentListTable.tsx
```diff
@@ -153,15 +161,16 @@ const ContentListTable = () => {
...
+                            distribution_versions,
                             account_id,
                             org_id
                         }) => (
                             <Tr key={ uuid }>
                                 <Td>{name}</Td>
                                 <Td>{url}</Td>
                                 <Td>{distribution_arch}</Td>
-                                <Td>{distribution_version}</Td>
+                                <Td>{versionDisplay(distribution_versions)}
```
jlsherrill, line 172: if there's a better way to do this than a function, let me know!

Andrewgdewar: No, that's totally fine. 
We may want to move the "versionDisplay" method later to a global helper. As we may want to use it in the "Edit" or other views. 
But for now it's okay here.

Andrewgdewar (approved):
LGTM

PR #9, opened 2022-06-24
Fixes 41: Add create content modal

Andrewgdewar:
To test this PR, one will need to have checked out [this backend PR](https://github.com/content-services/content-sources-backend/pull/59) (or have it merged to master).

This adds the bulk create modal.
This adds the following required API's: 
- Arch/Version repo
- Validate Content
- Bulk Create

Additional changes: 
- Added empty state/emptyTable state/Loading table state with skeleton
- Worked entirely too much on gpgKey integration (80% functionality)
- Hid said GPG key functionality.
- Replaced @rverdile's react Context code (sorry!) to avoid [this bug](https://github.com/patternfly/patternfly-react/issues/7761)
- Updated Eslint rules
- Updated some cloud-services npm modules to latest
- Added formik/Yup node modules
- Fixed an issue with optimistic updates.
- Added "Add 20" magic button within the create modal.
- Fixed issue when deleting last item on a page > 1, where the page wouldn't navigate the user back.
- Fixed issue with page not resetting when changing the search filters

Still needs: 
- <s>Add readable specific field errors if bulk create fails (currently just throws generic), waiting on [this ticket](https://issues.redhat.com/browse/HMSCONTENT-133)</s>

- Direction on what to do when we have an url that fails it's head request (currently prevents continuing): 
Screen Shot 2022-08-04 at 2 24 44 PM

Note on the above (not sure if this is considered a bug), pinging an address that has an http redirect will return success.
Example: https://quay.io/stuffandthings/thatprobablydontexist


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-41

src/services/Content/ContentApi.ts
```diff
@@ -44,3 +54,8 @@ export const deleteContentListItem: (uuid: string) => Promise<void> = async (uui
   const { data } = await axios.delete(`/api/content_sources/v1/repositories/${uuid}`);
   return data;
 };
+
+export const getRepositoryParams: () => Promise<RepositoryParamsResponse> = async () => {
+  const { data } = await axios.get('/api/content_sources/v1/repository_parameters/');
+  return data;
+};
```
Andrewgdewar, line 61: Made similar changes to your PR: @rverdile

src/services/Content/ContentQueries.ts
```diff
@@ -79,3 +84,9 @@ export const useDeleteContentItemMutate = (
     },
   });
 };
+
+export const useRepositoryParams = () =>
+  useQuery<RepositoryParamsResponse>(REPOSITORY_PARAMS_KEY, getRepositoryParams, {
+    keepPreviousData: true,
+    staleTime: Infinity,
+  });
```
Andrewgdewar, line 92: We need to mark the datas staletime as: Infinity... this will ensure we don't refetch it.

src/components/ContentListTable/ContentListTable.tsx
```diff
@@ -45,14 +64,20 @@ const useStyles = createUseStyles({
...
   const [perPage, setPerPage] = useState(storedPerPage);
 
   const hasActionPermissions = true; // TODO: Incorporate permissions here later.
 
+  const {
+    isLoading: repositoryParamsLoading,
+    error: repositoryParamsError,
+    isError: repositoryParamsIsError,
+    // data: repositoryParamsData,
+  } = useRepositoryParams();
```
Andrewgdewar, line 79: I added this call within the main listTable so that I can access the error/isError/isLoading states and merge them with the other loading/error states below.

src/components/AddContent/AddContent.tsx
```diff
@@ -0,0 +1,479 @@
...
+  });
+  const queryClient = useQueryClient();
+
+  const repositoryParams =
+    queryClient.getQueryData<RepositoryParamsResponse>(REPOSITORY_PARAMS_KEY);
+
+  console.log(
+    '🚀 ~ file: AddContent.tsx ~ line 142 ~ AddContent ~ repositoryParams',
+    repositoryParams,
+  );
```
Andrewgdewar, line 147: Here I access the data via queryClient.getQueryData, 
It's already been called on the main page, I now just need to use the key to get the data again.

src/services/Content/ContentQueries.ts
```diff
@@ -3,20 +3,24 @@ import { QueryClient, useMutation, useQuery } from 'react-query';
...
 export const CONTENT_LIST_KEY = 'CONTENT_LIST_KEY';
-export const CONTENT_REPO_PARAMS_KEY = 'CONTENT_REPO_PARAMS_KEY';
+export const REPOSITORY_PARAMS_KEY = 'REPOSITORY_PARAMS_KEY';
+export const CREATE_PARAMS_KEY = 'CREATE_PARAMS_KEY';
 
-export const useContentListQuery = (page: number, limit: number, filterData: filterData) =>
+export const useContentListQuery = (page: number, limit: number, filterData: FilterData) =>
   useQuery<ContentListResponse>(
-    [CONTENT_LIST_KEY, page, limit, filterData],
+    [CONTENT_LIST_KEY, page, limit, ...Object.values(filterData)],
```
Andrewgdewar, line 23: @rverdile In order to track these values I needed to flatten out the filterData object to get the key to work. As with the current filterData, I think it was being detected as when the comparison is run, I was seeing the key print as Object[Object].

jlsherrill:
I'm guessing this appeared after hiding the gpg key section, but now the dropdown  for versions clips the end of the  section and you can't see it:

image

jlsherrill:
One issue:

1.  enter a url without any version info in it
2.  select some set of versions
3. go back to the url and change it slightly
4. versions change back to 'any version'
I thought we had mentioned this and you said it shouldn't do this, but i may be misremembering, i know you said it was tricky

jlsherrill:
seeing a weird format issue when trying to add a repo and the versions were auto-selected from the url.  The structure of the distribution_versions is wrong:

```
[
    {
        "name": "test4",
        "url": "http://yum.theforeman.org/katello/4.2/katello/el7/x86_64/",
        "distribution_arch": "x86_64",
        "distribution_versions": [
            [
                "el7"
            ]
        ],
        "gpgKey": ""
    },
    {
        "name": "test8",
        "url": "http://mirror.centos.org/centos/8-strea/BaseOS/x86_64/os/",
        "distribution_arch": "x86_64",
        "distribution_versions": [
            "el9"
        ],
        "gpgKey": ""
    }
]
```

The first object in this list was auto-selected, the 2nd one was not.  Notice that there is an array within an array in the first object which isn't valid

Andrewgdewar:
> seeing a weird format issue when trying to add a repo and the versions were auto-selected from the url. The structure of the distribution_versions is wrong:
> 
> ```
> [
>     {
>         "name": "test4",
>         "url": "http://yum.theforeman.org/katello/4.2/katello/el7/x86_64/",
>         "distribution_arch": "x86_64",
>         "distribution_versions": [
>             [
>                 "el7"
>             ]
>         ],
>         "gpgKey": ""
>     },
>     {
>         "name": "test8",
>         "url": "http://mirror.centos.org/centos/8-strea/BaseOS/x86_64/os/",
>         "distribution_arch": "x86_64",
>         "distribution_versions": [
>             "el9"
>         ],
>         "gpgKey": ""
>     }
> ]
> ```
> 
> The first object in this list was auto-selected, the 2nd one was not. Notice that there is an array within an array in the first object which isn't valid

Working on this now. I attempted an incomplete fix yesterday.
Screen Shot 2022-08-09 at 8 21 09 AM

src/services/Content/ContentQueries.ts
```diff
@@ -25,15 +29,49 @@ export const useContentListQuery = (page: number, limit: number, filterData: fil
...
-export const useContentRepoParamsQuery = () =>
-  useQuery<ContentRepoParamsResponse>([CONTENT_REPO_PARAMS_KEY], () => getContentRepoParams());
+export const useAddContentQuery = (queryClient: QueryClient, request: CreateContentRequest) => {
+  const { notify } = useNotification();
+  return useMutation(() => AddContentListItems(request), {
+    onSuccess: () => {
+      notify({
+        variant: AlertVariant.success,
+        title: `Successfully added ${request.length} ${request.length > 1 ? 'items' : 'item'}.`,
+        // description: error?.message,
```
jlsherrill, line 39: intended?

rverdile (commented):
This should probably be its own issue, but the call to list API using the "version" and "arch" params, instead of "available_for_version" and "available_for_arch". This means, for example, if you filter by "el7" you only get "el7" repos instead of "el7" and "any".

 I forgot we even had two different options :)

rverdile (commented):
There's a "lorem ipsum" help popover next to the "x" button on the create modal. Not sure if you left that there on purpose, so just pointing it out.

rverdile (commented):
was it intended to keep the "too long" and "too short" validation or was that just a proof of concept?

src/components/AddContent/components/ContentValidity.tsx
```diff
@@ -0,0 +1,71 @@
...
+    name: string;
+    url: string;
+    gpgKey: string;
+    gpgLoading: boolean;
+    expanded: boolean;
+  }>;
+  loading?: boolean;
+}
+
+const ContentValidity = ({ touched, errors, loading }: Props) => {
```
rverdile, line 32: If I'm creating with a URL where I get "no such host", should it block me from creating or just give me a warning? I'm not sure what causes that error, but what if the URL is just down momentarily and I know it's valid?

rverdile: This section of the code it not necessarily relevant, just clicked here to get the conversation UI

src/services/Content/ContentApi.ts
```diff
@@ -30,21 +40,42 @@ export interface ContentListResponse {
...
+  label: string;
+};
+
+export type FilterData = {
   searchQuery: string;
   versions: Array<string>;
   arches: Array<string>;
 };
 
+export type ValidateItem = {
```
rverdile, line 59: what's the "Item" here in ValidateItem?

src/components/AddContent/components/ContentValidity.tsx
Andrewgdewar: That's what I meant by this piece, I think the general consensus is that we want to stop the users (as per design feedback thus far).
Screen Shot 2022-08-15 at 2 18 23 PM

src/services/Content/ContentApi.ts
Andrewgdewar: I suppose it would be "object" or "type", not fussy, can change the name. Thoughts? or perhaps "ValidationItem"?

src/components/AddContent/components/ContentValidity.tsx
Andrewgdewar: > was it intended to keep the "too long" and "too short" validation or was that just a proof of concept?

I think the too long/too short make sense, I just need feedback on limits there. Logical limits of between 2 - 50 are currently set.

rverdile: You can leave everything as-is. I was just pointing some things out in case they were being overlooked, but it sounds like they're not.

src/services/Content/ContentApi.ts
rverdile: Item is okay. I understand what it is now.

rverdile (commented):
I'll leave final approval for justin

jlsherrill (approved):
ack

PR #10, opened 2022-07-01
Fixes 79: Search and filter content sources list

rverdile:
**Adds:** 
- `ContentListFilters.tsx`
- `ContentListChips.tsx`
- `DropdownSelect.tsx`
- ~~`ContentListToolbar.tsx`~~

ContentListFilters, ContentListChips, and DropdownSelect all work nicely together with the existing table to make the search bar with filters. This is what I had originally (and what ContentListTable currently renders).

 ~~_Then_, I found the [patternfly Toolbar component](https://www.patternfly.org/v4/components/toolbar/), which is maybe what I should've used, so I tried implementing the same thing with that. But, that implementation is still buggy and seems a lot messier than the first one.~~ Not using this.

Still needs
- [x] ~~Fix ContentListToolbar?~~
- [x] API call to repository_params endpoint (right now uses mocked data)

### DIFF
--- a/src/components/ContentListTable/ContentListChips.tsx
+++ b/src/components/ContentListTable/ContentListChips.tsx
@@ -1,0 +1,60 @@
+import React, { useContext } from 'react';
+import { Button, Chip, ChipGroup } from '@patternfly/react-core';
+import { ContentListContext } from './ContentListContext';
+
+const ContentListChips = () => {
+
+    const {
+      filterData,
+      searchQuery,
+      setSearchQuery,
+      selectedVersions,
+      setSelectedVersions,
+      selectedArches,
+      setSelectedArches,
+    } = useContext(ContentListContext)
+
+    const deleteItem = (id: string, chips, setChips) => {
+        const copyOfChips = [...chips];
+        const filteredCopy = copyOfChips.filter(chip => chip !== id);
+        setChips(filteredCopy);
+    };
+
+    const clearFilters = () => {
+        setSelectedVersions([])
+        setSelectedArches([])
+        setSearchQuery('')
+    }
+
+    return (
+        <>
+            <ChipGroup categoryName={'Version'}>
+                {selectedVersions.map(version => (
+                    <Chip key={version} onClick={() => deleteItem(version, selectedVersions, setSelectedVersions)}>
+                        {version}
+                    </Chip>
+                ))}
+            </ChipGroup>
+            <ChipGroup categoryName={'Architecture'}>
+                {selectedArches.map(arch => (
+                    <Chip key={arch} onClick={() => deleteItem(arch, selectedArches, setSelectedArches)}>
+                        {arch}
+                    </Chip>
+                ))}
+            </ChipGroup>
+                {(searchQuery !== '') && <ChipGroup categoryName={'Name/URL'}>
+                    <Chip key={'search_chip'} onClick={() => setSearchQuery('')}>
+                        {searchQuery}
+                    </Chip>
+            </ChipGroup>}
+            {(filterData.searchQuery !== ''
+                && searchQuery !== ''
+                || !!selectedVersions?.length
+                || !!selectedArches?.length
+            )
+            && <Button onClick={clearFilters} variant="link" isInline>Clear filters</Button>}
+        </>
+    );
+}
+
+export default ContentListChips
--- a/src/components/ContentListTable/ContentListChips.tsx
+++ b/src/components/ContentListTable/ContentListChips.tsx
@@ -1,60 +1,68 @@
 import React, { useContext } from 'react';
 import { Button, Chip, ChipGroup } from '@patternfly/react-core';
 import { ContentListContext } from './ContentListContext';
+import Hide from '../Hide/Hide';
 
 const ContentListChips = () => {
+  const {
+    filterData,
+    searchQuery,
+    setSearchQuery,
+    selectedVersions,
+    setSelectedVersions,
+    selectedArches,
+    setSelectedArches,
+  } = useContext(ContentListContext);
 
-    const {
-      filterData,
-      searchQuery,
-      setSearchQuery,
-      selectedVersions,
-      setSelectedVersions,
-      selectedArches,
-      setSelectedArches,
-    } = useContext(ContentListContext)
+  const deleteItem = (id: string, chips, setChips) => {
+    const copyOfChips = [...chips];
+    const filteredCopy = copyOfChips.filter((chip) => chip !== id);
+    setChips(filteredCopy);
+  };
 
-    const deleteItem = (id: string, chips, setChips) => {
-        const copyOfChips = [...chips];
-        const filteredCopy = copyOfChips.filter(chip => chip !== id);
-        setChips(filteredCopy);
-    };
+  const clearFilters = () => {
+    setSelectedVersions([]);
+    setSelectedArches([]);
+    setSearchQuery('');
+  };
 
-    const clearFilters = () => {
-        setSelectedVersions([])
-        setSelectedArches([])
-        setSearchQuery('')
-    }
+  return (
+    <>
+      <Hide hide={searchQuery == '' && !selectedVersions?.length && !selectedArches?.length}>
+        <ChipGroup categoryName={'Version'}>
+          {selectedVersions.map((version) => (
+            <Chip
+              key={version}
+              onClick={() => deleteItem(version, selectedVersions, setSelectedVersions)}
+            >
+              {version}
+            </Chip>
+          ))}
+        </ChipGroup>
+        <ChipGroup categoryName={'Architecture'}>
+          {selectedArches.map((arch) => (
+            <Chip key={arch} onClick={() => deleteItem(arch, selectedArches, setSelectedArches)}>
+              {arch}
+            </Chip>
+          ))}
+        </ChipGroup>
+        {searchQuery !== '' && (
+          <ChipGroup categoryName={'Name/URL'}>
+            <Chip key={'search_chip'} onClick={() => setSearchQuery('')}>
+              {searchQuery}
+            </Chip>
+          </ChipGroup>
+        )}
+        {((filterData.searchQuery !== '' && searchQuery !== '') ||
+          !!selectedVersions?.length ||
+          !!selectedArches?.length) && (
+          <Button onClick={clearFilters} variant='link' isInline>
+            Clear filters
+          </Button>
+        )}
+      </Hide>
+    </>
+  );
+};
 
-    return (
-        <>
-            <ChipGroup categoryName={'Version'}>
-                {selectedVersions.map(version => (
-                    <Chip key={version} onClick={() => deleteItem(version, selectedVersions, setSelectedVersions)}>
-                        {version}
-                    </Chip>
-                ))}
-            </ChipGroup>
-            <ChipGroup categoryName={'Architecture'}>
-                {selectedArches.map(arch => (
-                    <Chip key={arch} onClick={() => deleteItem(arch, selectedArches, setSelectedArches)}>
-                        {arch}
-                    </Chip>
-                ))}
-            </ChipGroup>
-                {(searchQuery !== '') && <ChipGroup categoryName={'Name/URL'}>
-                    <Chip key={'search_chip'} onClick={() => setSearchQuery('')}>
-                        {searchQuery}
-                    </Chip>
-            </ChipGroup>}
-            {(filterData.searchQuery !== ''
-                && searchQuery !== ''
-                || !!selectedVersions?.length
-                || !!selectedArches?.length
-            )
-            && <Button onClick={clearFilters} variant="link" isInline>Clear filters</Button>}
-        </>
-    );
-}
-
-export default ContentListChips
+export default ContentListChips;
--- a/src/components/ContentListTable/ContentListContext.tsx
+++ b/src/components/ContentListTable/ContentListContext.tsx
@@ -1,0 +1,120 @@
+import { createContext, Dispatch, SetStateAction, useEffect, useState } from 'react';
+import useDebounce from '../../services/useDebounce';
+import { useContentRepoParamsQuery } from '../../services/Content/ContentQueries';
+import { filterData } from '../../services/Content/ContentApi';
+
+
+type ContentListContextProviderProps = {
+  children: React.ReactNode
+}
+
+interface RepoParams {
+  distribution_versions: Array<{ name: string, label: string }>
+  distribution_arches: Array<{ name:string, label: string }>
+}
+
+export type Filters = 'Name/URL' | 'Version' | 'Architecture'
+
+interface Context {
+  searchQuery: string,
+  setSearchQuery: Dispatch<SetStateAction<string>>,
+  filterType: Filters,
+  setFilterType: Dispatch<SetStateAction<Filters>>,
+  selectedVersions: Array<string>,
+  debouncedSelectedVersions: Array<string>,
+  setSelectedVersions: Dispatch<SetStateAction<Array<string>>>,
+  selectedArches: Array<string>,
+  debouncedSelectedArches: Array<string>
+  setSelectedArches: Dispatch<SetStateAction<Array<string>>>,
+  repoParams: RepoParams,
+  filterData: filterData,
+}
+
+export const ContentListContext = createContext<Context>({
+  searchQuery: '',
+  setSearchQuery: () => null,
+  filterType: 'Name/URL',
+  setFilterType: () => null,
+  selectedVersions: [],
+  debouncedSelectedVersions: [],
+  setSelectedVersions: () => null,
+  selectedArches: [],
+  debouncedSelectedArches: [],
+  setSelectedArches: () => null,
+  repoParams: { distribution_versions: [], distribution_arches: [] },
+  filterData: { searchQuery: '', versions: [], arches: [] },
+});
+
+const ContentListContextProvider = ({
+  children
+}: ContentListContextProviderProps) => {
+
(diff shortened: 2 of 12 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-79

src/components/ContentListTable/ContentListTable.tsx
```diff
@@ -131,6 +181,17 @@ const ContentListTable = () => {
...
       </Flex>
+      <Flex className={classes.chips}>
+        <ContentListChips
+            searchQuery={searchQuery}
+            debouncedSearchQuery={debouncedSearchQuery}
+            setSearchQuery={setSearchQuery}
+            selectedVersions={selectedVersions}
+            setSelectedVersions={setSelectedVersions}
+            selectedArches={selectedArchitectures}
+            setSelectedArches={setSelectedArchitectures}
```
Andrewgdewar, line 192: Make these variables internal state using react context within this component.
Do not pass down setters/variables to children, use react context to get them within the child components instead. 
I did something similar in katello's bulkDeleteModal, but the idea is that you only reveal the logic needed for the parent component, in this case the setSearchQuery variable should be all you need.

src/components/ContentListToolbar/ContentListToolbar.tsx
```diff
@@ -0,0 +1,208 @@
...
+    const [selectedFilters, setSelectedFilters] = useState({
+        filterType: 'Name/URL',
+        versions: [],
+        architectures: [],
+    });
+    const [filterIsOpen, setFilterIsOpen] = useState(false);
+    const [versionIsOpen, setVersionIsOpen] = useState(false);
+    const [archIsOpen, setArchIsOpen] = useState(false);
+
+    const selectFrom = (options: Array<string>) => {
```
Andrewgdewar, line 57: where possible, for this and all other simple functions, use a direct return. 
someFunc = (vars) => vars.dothings()

src/components/ContentListTable/ContentListChips.tsx
```diff
@@ -0,0 +1,67 @@
...
+    setSelectedArches,
+}: ContentListChipsProps) => {
+
+    const deleteItem = (id: string, chips, setChips) => {
+        const copyOfChips = [...chips];
+        const filteredCopy = copyOfChips.filter(chip => chip !== id);
+        setChips(filteredCopy);
+    };
+
+    const onClick = () => {
```
Andrewgdewar, line 30: A more meaningful name here would be clearFilters

rverdile:
You should know, on the backend main branch, there's a bug with filtering list by URL.

URL moved from `repository_configurations` to `repostories`, which causes a "URL column not found" error. A temporary work around is to remove the URL searching from this line and just search by name.

https://github.com/content-services/content-sources-backend/blob/main/pkg/dao/repository.go#L93

src/components/ContentListTable/ContentListTable.tsx
```diff
@@ -8,35 +8,66 @@ import {
...
+]
+}
 
 const useStyles = createUseStyles({
   actionContainer: {
     backgroundColor: global_BackgroundColor_100.value,
     justifyContent: 'space-between',
     padding: '16px 24px',
   },
+  chips: {
```
Andrewgdewar, line 66: Maybe rename to chipsContainer**

src/components/ContentListTable/ContentListTable.tsx
```diff
@@ -51,15 +82,37 @@ const ContentListTable = () => {
   const [page, setPage] = useState(1);
   const [perPage, setPerPage] = useState(storedPerPage);
 
+  const context = useContext(ContentListContext)
```
Andrewgdewar, line 85: de-structure here.

src/components/ContentListTable/ContentListTable.tsx
```diff
@@ -51,15 +46,43 @@ const ContentListTable = () => {
...
+    names.forEach(name => {
+          const found = namesLabels.find(v => v.name === name )
+          if(found) {
+            labels.push(found.label)
+          }
+        }
+    );
+    return labels
+  };
+
```
Andrewgdewar, line 70: Could this be done in context? As the labels are consumed here, the logic repoParams and versions/arches can take place in context as well.

src/components/ContentListTable/ContentListTable.tsx
```diff
@@ -108,82 +131,80 @@ const ContentListTable = () => {
...
-            variant={PaginationVariant.bottom}
-            onPerPageSelect={onPerPageSelect}
-          />
-        </FlexItem>
-      </Flex>
-    </Grid>
+      <Grid>
+        <Flex className={classes.actionContainer}>
+          <FlexItem>
+              <ContentListFilters/>
```
Andrewgdewar, line 137: 👍

src/components/ContentListTable/ContentListTable.tsx
```diff
@@ -108,82 +131,80 @@ const ContentListTable = () => {
...
+              page={page}
+              onSetPage={onSetPage}
+              widgetId='pagination-options-menu-top'
+              isCompact
+              onPerPageSelect={onPerPageSelect}
+            />
+          </FlexItem>
+        </Flex>
+        <Flex className={classes.chipsContainer}>
+          <ContentListChips/>
```
Andrewgdewar, line 153: 👍

src/components/ContentListTable/ContentListFilters.tsx
```diff
@@ -0,0 +1,62 @@
...
+import DropdownSelect from '../DropdownSelect/DropdownSelect';
+import { FilterIcon } from '@patternfly/react-icons';
+import { ContentListContext } from './ContentListContext';
+
+const ContentListFilters = () => {
+    const filters= ['Name/URL', 'Version', 'Architecture'];
+    const versions = {};
+    const arches = {};
+
+    const context = useContext(ContentListContext)
```
Andrewgdewar, line 12: destructure where able.

src/components/ContentListTable/ContentListContext.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  const [filterType, setFilterType] = useState<string>('Name/URL');
+  const [searchQuery, setSearchQuery] = useState('');
+  const [selectedVersions, setSelectedVersions] = useState<string[]>([]);
+  const [selectedArches, setSelectedArches] = useState<string[]>([]);
+
+  const filters = {
+    searchQuery, selectedVersions, selectedArches
+  }
+  const debouncedFilters = useDebounce(filters);
```
Andrewgdewar, line 52: const {searchQuery: debouncedSearchQuery, selectedVersions: debouncedSelectedVersions, etc... } = useDebounce(filters)

src/services/useDebouce.tsx
```diff
@@ -0,0 +1,17 @@
+import { useEffect, useState } from 'react'
```
Andrewgdewar, line 1: Change the name of this file (typo)

src/components/ContentListTable/ContentListTable.tsx
rverdile: I'm a little unclear on what you mean. Are you saying to move this function to context and just call it in ContentListTable?

Andrewgdewar: No, I mean as you only consume these variables here: 
debouncedSearchQuery,
    debouncedSelectedVersions,
    debouncedSelectedArches, 
 
 .. I would suggest placing your function inside context, and have the "Labels" used in this component. 
 So the labels would be passed down by context.

rverdile: good idea! thanks

rverdile:
I think I've addressed all the comments now

src/components/ContentListTable/ContentListChips.tsx
```diff
@@ -0,0 +1,52 @@
+import React, { useContext } from 'react';
+import { Button, Chip, ChipGroup } from '@patternfly/react-core';
+import { ContentListContext } from './ContentListContext';
+
+const ContentListChips = () => {
+
+    const context = useContext(ContentListContext)
```
Andrewgdewar, line 7: destructure

src/components/ContentListTable/ContentListChips.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+    const clearFilters = () => {
+        context.setSelectedVersions([])
+        context.setSelectedArches([])
+        context.setSearchQuery('')
+    }
+
+    return (
+        <>
+            <ChipGroup categoryName={'Version'}>
```
Andrewgdewar, line 23: Where there are strings as variables you can do ='Version'

src/components/ContentListTable/ContentListChips.tsx
```diff
@@ -0,0 +1,52 @@
...
+            </ChipGroup>
+                {(context.searchQuery !== '') && <ChipGroup categoryName={'Name/URL'}>
+                    <Chip key={'search_chip'} onClick={() => context.setSearchQuery('')}>
+                        {context.searchQuery}
+                    </Chip>
+            </ChipGroup>}
+            {(context.debouncedSearchQuery !== ''
+                && context.searchQuery !== ''
+                || context.selectedVersions.length !== 0
+                || context.selectedArches.length !== 0
```
Andrewgdewar, line 45: you can use the double bang here to say not 0, as javascript considers 0 a falsey value. So: 
!!context.selectedArches?.length

src/components/ContentListTable/ContentListContext.tsx
```diff
@@ -0,0 +1,121 @@
...
+        const found = namesLabels.find(v => v.name === name )
+        if(found) {
+          labels.push(found.label)
+        }
+      }
+    );
+    return labels
+  };
+
+  const labels = {
```
Andrewgdewar, line 91: 👍

src/components/ContentListTable/ContentListContext.tsx
```diff
@@ -0,0 +1,121 @@
...
+  } = useDebounce(filters)
+
+  const {
+    isLoading,
+    error,
+    isError,
+    data = { distribution_arches: [], distribution_versions: [] },
+  } = useContentRepoParamsQuery();
+
+  const getLabels = (type: string, names: Array<string>) => {
```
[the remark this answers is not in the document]
Andrewgdewar: If you do go the route of using a useCallback, I'd suggest returning an object containing both of the labels you need.

src/components/ContentListTable/ContentListContext.tsx
```diff
@@ -0,0 +1,121 @@
...
+  };
+
+  const labels = {
+    versions: getLabels('version', debouncedSelectedVersions),
+    arches: getLabels('arch', debouncedSelectedArches)
+  }
+
+  if (isError) throw error
+
+  if (isLoading) return <></>;
```
Andrewgdewar, line 98: I have the loading logic built out on my branch, when this is merged I'll likely need to do something here.
```suggestion
  if (isLoading) return <></>; // TODO: Implement loader to match the top-level loader here.
```

src/components/ContentListTable/ContentListFilters.tsx
```diff
@@ -0,0 +1,72 @@
...
+import DropdownSelect from '../DropdownSelect/DropdownSelect';
+import { FilterIcon } from '@patternfly/react-icons';
+import { ContentListContext } from './ContentListContext';
+
+const ContentListFilters = () => {
+    const filters= ['Name/URL', 'Version', 'Architecture'];
+    const versions = {};
+    const arches = {};
+
+    const {
```
Andrewgdewar, line 12: 👍

src/components/ContentListTable/ContentListFilters.tsx
```diff
@@ -0,0 +1,72 @@
...
+        setSearchQuery,
+        filterType,
+        setFilterType,
+        selectedVersions,
+        setSelectedVersions,
+        selectedArches,
+        setSelectedArches,
+    } = useContext(ContentListContext)
+
+    repoParams['distribution_arches'].forEach(arch => arches[arch.name] = arch.label );
```
Andrewgdewar, line 23: you'll want to to throw this into a useEffect with dependencies preventing it running every render.

src/components/ContentListTable/ContentListTable.tsx
```diff
@@ -51,6 +46,11 @@ const ContentListTable = () => {
   const [page, setPage] = useState(1);
   const [perPage, setPerPage] = useState(storedPerPage);
 
+  const {
+    debouncedSearchQuery,
+    labels
```
Andrewgdewar, line 51:
```suggestion
   filterData
```

Andrewgdewar: Screen Shot 2022-07-19 at 10 04 45 AM

Could we change this to filterData? As this is only used for the API we can keep the idea of it similar to within the API.

src/components/ContentListTable/ContentListChips.tsx
Andrewgdewar: Don't worry about this, I added a lint rule to fix/cover it.

jlsherrill:
https://issues.redhat.com/browse/14676876

src/components/ContentListTable/ContentListFilters.tsx
```diff
@@ -0,0 +1,72 @@
...
+    repoParams['distribution_versions'].forEach(version => versions[version.name] = version.label);
+
+    const onFilterChange = (): ReactElement => {
```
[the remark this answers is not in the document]
rverdile: trying to do this, but changing this to a component causes a error "Can't perform a React state update on an unmounted component"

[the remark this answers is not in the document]
Andrewgdewar: hmm.. 
Will look with you.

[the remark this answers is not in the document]
rverdile: just renamed instead like discussed

Andrewgdewar:
Screen Shot 2022-07-22 at 12 49 51 PM
Screen Shot 2022-07-22 at 12 50 36 PM

Andrewgdewar:
Screen Shot 2022-07-22 at 12 52 14 PM
Screen Shot 2022-07-22 at 12 52 34 PM

src/components/ContentListTable/ContentListTable.tsx
```diff
@@ -8,35 +8,30 @@ import {
...
 
 const useStyles = createUseStyles({
   actionContainer: {
     backgroundColor: global_BackgroundColor_100.value,
     justifyContent: 'space-between',
     padding: '16px 24px',
   },
+  chipsContainer: {
+    backgroundColor: global_BackgroundColor_100.value,
+    justifyContent: 'justifyContentFlexStart',
```
Andrewgdewar, line 32: does this work? I would think that the correct format would be flex-start here, as "justifyContentFlexStart" is specific to patternfly.

src/components/DropdownSelect/DropdownSelect.tsx
```diff
@@ -0,0 +1,55 @@
+import { Select, SelectOption, SelectVariant } from '@patternfly/react-core';
+import React, { Dispatch, SetStateAction, useState } from 'react';
+
+type DropdownSelectProps = {
+    options: Array<string>,
+    variant: SelectVariant.single | SelectVariant.checkbox
+    selectedProp: string | Array<string>,
+    setSelected: Dispatch<SetStateAction<any>>, // TODO type this
```
Andrewgdewar, line 8: Your type will likely look something like this: (type: 'Name/URL' | 'Version' | 'Architecture') => void;

rverdile: I'm not sure this works. The issue here is I wanted to be able to set a string (or filter in this case) or Array<string> (for versions/arches). The "single" variant would be setting a string, the "checkbox" variant would be setting an array.

src/components/ContentListTable/ContentListTable.tsx
rverdile: i changed it to 'flex-start' and it still looks okay, so I'll leave it like that. I don't remember where "justifyContentFlexStart" came from.

src/components/DropdownSelect/DropdownSelect.tsx
Andrewgdewar: Then just add both as the types with an "|" or operator?

rverdile: It doesn't seem to be as simple as that. I pushed up what I tried, but I can't seem to find a way to type this that doesn't complain.

`setSelected: Dispatch<SetStateAction<Array<string>>> | Dispatch<SetStateAction<Filters>>`

Andrewgdewar: setSelected: Dispatch<SetStateAction<Filters | Array<string>>>;

Andrewgdewar: After further attempts at fixing this, I am now onboard for the "any" method within the DropdownSelect. 😅

src/components/ContentListTable/ContentListTable.tsx
```diff
@@ -131,6 +136,11 @@ const ContentListTable = () => {
           />
         </FlexItem>
       </Flex>
+      <Hide hide={!selectedVersions.length && !selectedArches.length && searchQuery == ''}>
```
Andrewgdewar, line 139:
```suggestion
      <Hide hide={!showListChips}>
```

Andrewgdewar (approved):
Looking good, last comment was a bit #nit feel free to merge as is or with an update.

PR #27, opened 2022-09-02
Fixes 112: Add #pkgs to the repository table

avisiedo:
- Update ContentItem interface to add package_count field.
- Update view to add the column with the package_count field.
- The column title should be `Packages`.
- Its position will be after `Version` column.

### DIFF
--- a/src/components/AddContent/helpers.test.ts
+++ b/src/components/AddContent/helpers.test.ts
@@ -24,7 +24,6 @@ it('mapFormikToAPIValues', () => {
       versions: ['el7'],
       gpgLoading: false,
       expanded: false,
-      package_count: 0,
     },
   ];
 
@@ -35,6 +34,7 @@ it('mapFormikToAPIValues', () => {
       distribution_arch: 'x86_64',
       distribution_versions: ['el7'],
       gpgKey: '',
+      package_count: 0,
     },
   ];
 
--- a/src/components/AddContent/helpers.test.ts
+++ b/src/components/AddContent/helpers.test.ts
@@ -34,7 +34,6 @@ it('mapFormikToAPIValues', () => {
       distribution_arch: 'x86_64',
       distribution_versions: ['el7'],
       gpgKey: '',
-      package_count: 0,
     },
   ];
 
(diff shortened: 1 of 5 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-112

src/components/ContentListTable/ContentListTable.tsx
```diff
@@ -210,11 +210,12 @@ const ContentListTable = () => {
             </Thead>
             <Tbody>
               {contentList.map(
-                ({ uuid, name, url, distribution_arch, distribution_versions }: ContentItem) => (
+                ({ uuid, name, url, distribution_arch, package_count, distribution_versions }: ContentItem) => (
```
Andrewgdewar, line 213: What value will you display if it is zero? 
"Unknown", 0, "N/A" ? 
The below is the screenshot, I think we may need design input here though. 
We'll likely want to have some N/A test and a hover popup that explains why.
Screen Shot 2022-09-06 at 9 02 40 AM

src/components/AddContent/helpers.test.ts
```diff
@@ -34,6 +34,7 @@ it('mapFormikToAPIValues', () => {
       distribution_arch: 'x86_64',
       distribution_versions: ['el7'],
       gpgKey: '',
+      package_count: 0,
```
Andrewgdewar, line 37: not needed here

src/components/AddContent/helpers.ts
```diff
@@ -26,6 +26,7 @@ export const mapFormikToAPIValues = (
     name,
     url,
     distribution_arch: arch,
+    package_count: 0,
```
Andrewgdewar, line 29: not needed here.

src/components/ContentListTable/ContentListTable.tsx
Andrewgdewar: For now maybe leave it as "0" and we can address what do here at a later time.

Andrewgdewar (approved):
Ack!

PR #29, opened 2022-09-07
Fixes 161: Add introspection status column to list table

rverdile:
This depends on this backend PR: https://github.com/content-services/content-sources-backend/pull/95

What's left:
- [x] Tests
- [x] Not sure what the "Pending" status is supposed to look like since it's not on the mocks.


src/components/ContentListTable/ContentListTable.tsx
```diff
@@ -35,6 +35,12 @@ import ContentListFilters from './ContentListFilters';
 import Hide from '../Hide/Hide';
 import EmptyTableState from './components/EmptyTableState';
 import { useQueryClient } from 'react-query';
+import StatusIcon from './components/StatusIcon';
+
+export const red = global_danger_color_100.value;
+export const green = global_success_color_100.value;
+export const yellow = '#F0AB00';
+export const brown = '#795600';
```
rverdile, line 43: I wanted to put all the styling stuff in `StatusIcon.tsx`, but calling createUseStyles there gives an error: "React has detected a change in the order of Hooks called by..."

src/components/ContentListTable/ContentListTable.tsx
```diff
@@ -35,6 +35,12 @@ import ContentListFilters from './ContentListFilters';
 import Hide from '../Hide/Hide';
 import EmptyTableState from './components/EmptyTableState';
 import { useQueryClient } from 'react-query';
+import StatusIcon from './components/StatusIcon';
+
+export const red = global_danger_color_100.value;
+export const green = global_success_color_100.value;
```
Andrewgdewar, line 41: if you must use this in multiple components, we should likely create a "helper.ts" file for this.

src/components/ContentListTable/ContentListTable.tsx
```diff
@@ -35,6 +35,12 @@ import ContentListFilters from './ContentListFilters';
...
+export const yellow = '#F0AB00';
```
Andrewgdewar, line 42: Can you not find these colors within [patternfly](https://www.patternfly.org/v4/guidelines/colors) that are matching this?

Andrewgdewar:
Before you get too far in this, we do have a requirement to sort by status on this table. If that doesn't require any changes in the backend (as in the data structures you are using) then there shouldn't be any issues. 

Just thought I'd mention it for planning.

src/components/ContentListTable/components/StatusIcon.tsx
```diff
@@ -0,0 +1,68 @@
...
+const StatusIcon = ({ status, error, classes }: Props) => {
+  switch(status) {
+    case 'Valid':
+      return (
+        <Flex alignContent={{ default: 'alignContentCenter' }} direction={{ default: 'row' }}>
+          <FlexItem spacer={{ default: 'spacerSm' }}>
+            <CheckCircleIcon color={green} />
+          </FlexItem>
+          <FlexItem>
+            <span className={classes.green}>Valid</span>
```
Andrewgdewar, line 20: I think a better way to do this would be to first create a "green" color variable in a helper.ts, and then a component named StatusText with the span and passed in color variable.

Andrewgdewar: Make your classes within that component.

src/components/ContentListTable/ContentListTable.tsx
rverdile: I got these values from patternfly, but I did not find a global variable for them like red and green

rverdile:
updated

Andrewgdewar:
Looks good, you may want to rebase from my edit PR, as there are some changes to classes that you may need to update.

Andrewgdewar:
May want to rebase from master

Andrewgdewar:
As for the pending status. Default statuses that are not bad or good, are usually represented with the "info" color. 
So you may want to try playing with those here: https://www.patternfly.org/v4/guidelines/colors
Also if you need another icon, the info one makes sense to me in this case as well.

src/components/ContentListTable/components/StatusIcon.test.tsx
```diff
@@ -0,0 +1,43 @@
+import { render } from '@testing-library/react';
+import StatusIcon from './StatusIcon';
+
+it('Render with Pending status', () => {
+  const { queryByText } = render(
+    <StatusIcon
+      status='Pending'
+    />,
+  );
```
rverdile, line 9: I was hoping to query for the actual icon here but I can't find a way to do that

Andrewgdewar: I think the text is enough here.

src/components/ContentListTable/components/StatusText.tsx
```diff
@@ -0,0 +1,29 @@
...
+interface Props {
+  color: string
+  children?: React.ReactNode
+}
+
+const StatusText = ({ color, children }: Props) => {
+  const classes = useStyles();
+
+  return (
+    <span className={classes[color]}>{children}</span>
```
Andrewgdewar, line 25: You'll likely want to make the color type above  a string selection like: "red" | "green" | etc..

src/components/EditContentModal/helpers.test.ts
```diff
@@ -38,6 +38,8 @@ it('mapToDefaultFormikValues', () => {
       package_count: 25,
       distribution_versions: ['version1', 'etc'],
       distribution_arch: 'stuffAndThings',
+      status: 'stuffAndThings',
```
Andrewgdewar, line 41: +1 for consistency in naming.

Andrewgdewar (approved):
👍

Andrewgdewar:
I am rebasing off of this, ping me if you make changes @rverdile

PR #34, opened 2022-09-19
Fixes 158: Update UI to match demo Mocks

Andrewgdewar:
Updates:
- First column content, Name/url
- Naming of the overall app.  Use "custom repositories" not "content sources"
- Update table styling to match: [design mocks](https://marvelapp.com/prototype/e1150c1/screen/88547707)

### DIFF
--- a/src/services/Content/ContentQueries.ts
+++ b/src/services/Content/ContentQueries.ts
@@ -40,14 +40,19 @@ export const useAddContentQuery = (queryClient: QueryClient, request: CreateCont
   const { notify } = useNotification();
   return useMutation(() => AddContentListItems(request), {
     onSuccess: () => {
+      const title =
+        request?.length > 1
+          ? `${request?.length} custom repositories added`
+          : `Custom repository "${request?.[0]?.name}" added`;
       notify({
         variant: AlertVariant.success,
-        title: `Successfully added ${request.length} ${request.length > 1 ? 'items' : 'item'}.`,
+        title,
+        description: 'Repository introspection may take some time',
       });
       queryClient.invalidateQueries(CONTENT_LIST_KEY);
     },
     onError: (err: { response?: { data: string | Array<{ error: string | null }> } }) => {
-      let description = 'An error occurred.';
+      let description = 'An error occurred';
 
       switch (typeof err?.response?.data) {
         case 'string':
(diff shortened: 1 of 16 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-158

rverdile:
mockup

I noted all the differences. Some of these are either intentional or because of missing features, but I noted everything for the sake of completion. I bolded the ones where I don't think this is the case.

1. Selectors on the left hand side of the repo
2. **The search icon is on the right in a box (vs on the left and not in a box).**
3. **The sentence ends with a period**
4. **The text in this pop up is entirely different ("Successfully added 1 item")**
5. Kebab (vs no kebab)
6. Package count is missing because the backend PR isn't ready. Shows a spinner here for no packages, but currently it's blank.
7. **Status is "In progress" with a spinner (vs "Pending" with an icon)**
8. Sort icon on columns

Andrewgdewar:
> mockup
> 
> I noted all the differences. Some of these are either intentional or because of missing features, but I noted everything for the sake of completion. I bolded the ones where I don't think this is the case.

> 2. **The search icon is on the right in a box (vs on the left and not in a box).**
> 3. **The sentence ends with a period**
> 4. **The text in this pop up is entirely different (vs "Successfully added 1 item")**
> 7. **Status is "In progress" with a spinner (vs "Pending" with an icon)**

I take it you mean to say that the above should match. 

> 2. **The search icon is on the right in a box (vs on the left and not in a box).**
> 4. **The text in this pop up is entirely different (vs "Successfully added 1 item")**
- Will update, good catch!

> 3. **The sentence ends with a period**
I don't think this is intentional, I've corrected a number of small issues in copy from Maria in the past around this.

> 7. **Status is "In progress" with a spinner (vs "Pending" with an icon)**

This feels kind of weird to me, I think we should likely poke Maria and get some design feedback here.

rverdile:
I'm sure exactly which things we want to match the mocks and which we're happy with, so I'm just pointing out the differences.

Andrewgdewar:
Screen Shot 2022-09-23 at 2 26 36 PM

src/services/Content/ContentQueries.ts
```diff
@@ -35,14 +35,19 @@ export const useAddContentQuery = (queryClient: QueryClient, request: CreateCont
...
     onSuccess: () => {
+      const title =
+        request?.length > 1
+          ? `${request?.length} custom repositories added`
+          : `Custom repository "${request?.[0]?.name}" added`;
       notify({
         variant: AlertVariant.success,
-        title: `Successfully added ${request.length} ${request.length > 1 ? 'items' : 'item'}.`,
+        title,
+        description: 'The validation and package introspection process may take some time',
```
rverdile, line 45: Thinking about this message, I'm not sure which validation process would take some time? I know the mock says this, I'm just wondering if that messaging is still accurate. Also, I'm not sure if you'd say "package introspection" instead of "repository introspection". Maybe @jlsherrill can chime in here.

rverdile:
This is a fault from my last PR, not yours, but if you could remove the `title=blue` from [here](https://github.com/content-services/content-sources-frontend/blob/main/src/components/ContentListTable/components/StatusIcon.tsx#L60), it would no longer say "blue" when hovering over the Pending icon.

src/services/Content/ContentQueries.ts
jlsherrill: yes, i'd say "repository introspection" and i also would not mention validation taking some time.

Andrewgdewar: Roger

PR #36, opened 2022-09-23
Fixes 77: add gpg key and verification option

rverdile:
- Unhides the GPG key form in the Edit and Add modals. 
- Adds a radio button for the metadata verification option.

I was thinking GPG Key and the verification option should be their own components, but it seems like that would be more than a little disruptive to the current formik stuff. I welcome suggestions there (and anywhere).

### DIFF
--- a/src/components/EditContentModal/EditContentModal.tsx
+++ b/src/components/EditContentModal/EditContentModal.tsx
@@ -5,12 +5,13 @@ import {
   FormGroup,
   Modal,
   ModalVariant,
-  Popover, Radio,
+  Popover,
+  Radio,
   SelectVariant,
   Stack,
   StackItem,
   TextInput,
-  Tooltip
+  Tooltip,
 } from '@patternfly/react-core';
 import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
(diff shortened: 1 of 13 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-77

src/components/EditContentModal/EditContentModal.tsx
```diff
@@ -5,12 +5,12 @@ import {
   FormGroup,
   Modal,
   ModalVariant,
-  Popover,
+  Popover, Radio,
```
Andrewgdewar, line 8: A linter addon would have caught this.

Andrewgdewar: I did a thing 😆

Andrewgdewar: It was format you needed to run in this case.

Andrewgdewar: I have prettier and eslint as addons for this reason. 
They work together to point out and automagically formatting issues. 
The only reason I'm a stickler is it makes later merging/rebasing easier.

rverdile: what's the command for that?

PR #38, opened 2022-10-05
Fixes 208: Add api validation for gpg-keys

Andrewgdewar:
This adds validation functionality to be paired with [this backend API ticket](https://github.com/content-services/content-sources-backend/pull/116), allowing the gpg_keys to: 

- Show errors when gpg_key is invalid
- Show errors when Package and metadata verification is set but gpg_key signature does not match.
- Auto-selection of Package and metadata verification option (based on url metadata_verification set to true)
- Auto-selection/disabling of metadata verification option (in case where url metadata_verification set to false).

- [x] Needs tests

### DIFF
--- a/src/components/AddContent/AddContent.test.tsx
+++ b/src/components/AddContent/AddContent.test.tsx
@@ -1,4 +1,4 @@
-import { act, fireEvent, render } from '@testing-library/react';
+import { act, fireEvent, render, waitFor } from '@testing-library/react';
 import {
   defaultValidationErrorData,
   passingValidationErrorData,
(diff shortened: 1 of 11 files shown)


src/components/AddContent/AddContent.test.tsx
```diff
@@ -119,6 +126,91 @@ it('expect "url" input to show a validation error', async () => {
...
+    button?.click();
+  });
+  const urlInput = queryByPlaceholderText('https://');
+  expect(urlInput).toBeInTheDocument();
+  if (urlInput) {
+    await act(async () => {
+      fireEvent.change(urlInput, { target: { value: 'https://bobTheBuilder.com' } });
+    });
+  }
+  expect(queryByText('Invalid URL')).not.toBeInTheDocument();
```
jlsherrill, line 200: i guess there's no way to reference the text via some constant rather than duplicating it?  I only say this because if the text changes, this test would still keep on passing even though its really not testing anything (and wouldn't detect a failure)

jlsherrill (approved):
One question, but not holding up testing

src/components/AddContent/AddContent.test.tsx
Andrewgdewar: Oh like a global constants for validation errors? We could do that.

jlsherrill: if it makes sense to do that more globally, lets file a ticket (i'm happy to do that)

mshriver:
https://issues.redhat.com/browse/HMSCONTENT-199 is going to be used to capture test coverage for this feature at a larger scope than this individual PR.

PR #52, opened 2022-10-19
Fixes 230: Fix UI redirect to /content

Andrewgdewar:
This change ensures that navigation will go to "/content" not "/content/anything-else". 

**Caveat:** 
When a selection is made on the external "Filter by tags" component found at the top of the page, it redirects our app to "/content/content#<whatever it adds>" instead of the now intentioned  "/content#<whatever it adds>" 

Screen Shot 2022-10-19 at 2 09 15 PM

Our redirect works only on actual navigation, the above uses a "replace" method to append parameters at the end of the URL.

Meaning this incorrect URL reference to be fixed, either within this component or where-ever the configured path is gotten from, likely another console dot configuration repo? @jlsherrill  

Additional changes: 
- Updated "react-router-dom"
- Removed 2 unused example components.


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-230

jlsherrill (approved):

jlsherrill:
i'm not sure where we would have configured it to do /content/content in any config, but i'll dig around

jlsherrill:
that said, i don't actually see the behavior you mention with the tags filter, it takes me to:  `https://stage.foo.redhat.com:1337/beta/insights/content#workloads=SAP&SIDs=&tags=`

swadeley:
In my EE test environment, when I navigate to the "Custom Repositories" page  I see `openshiftapps.com/insights/content/content`
In the stage environment, I did briefly see `/content/content`, but then I see `insights/content#SIDs=&tags=`

Andrewgdewar:
> In my EE test environment, when I navigate to the "Custom Repositories" page I see `openshiftapps.com/insights/content/content` In the stage environment, I did briefly see `/content/content`, but then I see `insights/content#SIDs=&tags=` EDIT: Click the "All apps " link, click "[Repositories]" while watching the URL (see it change to `/content/content` and then to `insights/content#SIDs=&tags=`)

I'm thinking there may be more work to do external to our app referencing our URL. 
This work is definitely needed though, so we may want to ack/merge this and then work on the (likely external) config in a separate ticket.

Andrewgdewar:
I added a change that seems to address the link issue (in stage at least).
Screen Shot 2022-10-26 at 11 54 16 AM

fec.config.js
```diff
@@ -3,7 +3,7 @@ module.exports = {
   useProxy: true,
   proxyVerbose: true,
   interceptChromeConfig: true,
-  appUrl: '/apps/content-sources',
+  appUrl: '/insights/content',
```
Andrewgdewar, line 6: This is said change.

PR #55, opened 2022-10-26
Fixes 244: Hide package verification

Andrewgdewar:
It may be a while before image builder supports metadata verification.  This disables the option in the UI.


Andrewgdewar:
This is now redundant, control of this feature will now be done through feature flags.

PR #57, opened 2022-10-28
Fixes 251: Add feature flag support

Andrewgdewar:
This adds support for "unleash" feature flags. 

To turn feature flags on and off go [here](https://insights-stage.unleash.devshift.net/projects/default).

The toggleable feature flag added for this PR is content-sources.hide-package-verification. 

The expected behaviour is as follows, open the add/edit modals and type something into the gpg-key field. 

While Enabled, the "Use GPG key for" input group should be invisible:
Screen Shot 2022-10-28 at 1 42 08 PM

While Disabled, the above field should be present in both modals.

**Note**: A refresh may be needed after toggling flags to ensure that fresh data has been pulled (I currently don't know what console dot has set the [refresh interval](https://docs.getunleash.io/sdks/proxy-react#initialization) to in the configuration of this service).

Additional Information:  An overview of the necessary changes for this PR can be found [here](http://front-end-docs-insights.apps.ocp4.prod.psi.redhat.com/feature-flags) and [here](http://front-end-docs-insights.apps.ocp4.prod.psi.redhat.com/feature-flags/managing-feature-flags).


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-239

src/middleware/AppContext.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { useFlag } from '@unleash/proxy-client-react';
+
+export interface AppContextInterface {
+  packageVerification: boolean;
+}
+
+export const AppContext = createContext({} as AppContextInterface);
+
+export const ContextProvider = ({ children }: { children: ReactNode }) => {
+  const packageVerification = useFlag('content-sources.package-verification') || false;
```
jlsherrill, line 11: would it make sense to turn this on by default?   what happens in ephemeral (i wouldn't want it to default to false i don't think)?

src/components/ContentListTable/components/AddContent/AddContent.tsx
```diff
@@ -272,6 +276,7 @@ const AddContent = ({ isLoading }: Props) => {
           ({ name, label }) => url.includes(name) || url.includes(label),
         )?.label;
         if (newVersion) versions = [newVersion];
+        if (isEmpty(versions)) versions = ['any'];
```
Andrewgdewar, line 279: This fixes a rare bug where the UI sends an empty string instead of "any" or ["any"] for arch/version. 
I know the API corrects this, but thought it better practice to enforce from ui as well.

jlsherrill (approved):

jlsherrill:
I think this is supposed to be associated with https://issues.redhat.com/browse/HMSCONTENT-251 and not 239?

PR #60, opened 2022-11-08
Refs 231: add /settings route

jlsherrill:
so that we can use the settings navigation


jlsherrill:
this adds a second route, which we will use in the navigation.  We could look at removing the first route, but i wanted to do this incrementally and not break all tests

jlsherrill:

Andrewgdewar (approved):
Much route, ACK!

jlsherrill:
https://issues.redhat.com/browse/HMS-363

PR #64, opened 2022-11-16
Fixes 263: show N/A for Invalid repositories

avisiedo:
- Update component to display N/A into the Packages column when the repository state is Invalid.
- Add unit tests for the component.

### DIFF
--- a/src/components/ContentListTable/components/PackageCount.tsx
+++ b/src/components/ContentListTable/components/PackageCount.tsx
@@ -32,6 +32,14 @@ const PackageCount = ({ rowData }: Props) => {
     );
   }
 
+  if (rowData.status === 'Invalid') {
+    return (
+      <Tooltip isContentLeftAligned content='Repository is invalid please review it'>
+        <Text className={classes.text}>N/A</Text>
+      </Tooltip>
+    )
+  }
+
   return (
     <>
       <Hide hide={!modalOpen}>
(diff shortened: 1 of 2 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-263

avisiedo:
E2E Test Scenario

> I have not found a way to set a repository invalid using the service in an usual way, so I trick the result by updating directly the database, so the below description is keeping that in mind

Given an application with one repository with valid options
   and updating the status field into the database for that repository to Invalid status value
When I display the table of the repositories
  Then I get the value for Packages equals to N/A
    and the status is Invalid
    and the value is not clickable

avisiedo:
thanks you @Andrewgdewar for helping me out to understand better unit testing in react.

Andrewgdewar (approved):
👍

src/components/ContentListTable/components/PackageCount.tsx
```diff
@@ -32,6 +32,14 @@ const PackageCount = ({ rowData }: Props) => {
     );
   }
 
+  if (rowData.status === 'Invalid') {
+    return (
+      <Tooltip isContentLeftAligned content='Repository is invalid please review it'>
```
swadeley, line 37:
```suggestion
      <Tooltip isContentLeftAligned content='Repository is invalid.'>
```
Better punctuation would be "Repository is invalid, please review it.' but style guides say not to use "please" so its best just to make the tool tip short and to the point.

PR #77, opened 2022-12-15
Fixes 287: Fix for null store error when navigating

Andrewgdewar:
Fixes a small issue with the redux store being "null" or undefined.

A fix on the main ReduxProvider was already present but there was the use of "Main" component which was imported from the shared front-end components, which needed to be removed as it was using a provider internally.

NOTE TO QE: your main page container id may have changed, I'm not sure if this will affect your existing tests.

### DIFF
--- a/src/AppEntry.tsx
+++ b/src/AppEntry.tsx
@@ -24,9 +24,8 @@ export default function AppEntry({ logger }: AppEntryProps) {
     resetStore();
     if (logger) {
       return createStore(logger).store;
-    } else {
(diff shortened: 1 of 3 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-287

mshriver (approved):

mshriver:
I'm going to address these changes in the QE repo along with the larger change to adding popular repos tab.

PR #80, opened 2022-12-21
Fixes 310: Adds polling for "Pending" repositories

Andrewgdewar:


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-310

jlsherrill:
Works great!  The only comment i have is that we should probably change the pending icon to something that indicates a polling action more.  Something like https://www.patternfly.org/v4/components/spinner

Andrewgdewar:
> Works great! The only comment i have is that we should probably change the pending icon to something that indicates a polling action more. Something like https://www.patternfly.org/v4/components/spinner

Should we also switch the text around pending? 
I thought there was alternate copy.. 
Screen Shot 2023-01-04 at 7 51 34 AM

jlsherrill:
yeah, I think that alternate copy looks good

jlsherrill (approved):

jlsherrill:
[test]

swadeley:
Works, looks good, thank you

PR #82, opened 2023-01-03
Fixes 318: Fix for Nan value upon deletion error 

Andrewgdewar:
Fix for Nan value on pagination component upon deleting an item from the list.

To recreate, simply have a list of greater than 10 items, set the pagination perPage value to 10 and delete 1 item on the first page. 
Looking at the console without this change you should see a react error around Nan values (if using a 3g network connection setting this is also viewable in the pagination component as "Nan")

### DIFF
--- a/src/services/Content/ContentQueries.ts
+++ b/src/services/Content/ContentQueries.ts
@@ -58,6 +58,7 @@ export const useContentListQuery = (
   const [pollCount, setPollCount] = useState(0);
 
   return useQuery<ContentListResponse>(
+    // Below MUST match the "contentListKeyArray" seen below in the useDeleteContent.
     [CONTENT_LIST_KEY, page, limit, sortBy, ...Object.values(filterData)],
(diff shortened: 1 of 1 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-318

src/services/Content/ContentQueries.ts
```diff
@@ -171,12 +171,13 @@ export const useDeleteContentItemMutate = (
   filterData?: FilterData,
   sortString?: string,
 ) => {
+  // Below MUST match the "useContentList" key found above or updates will fail.
```
jlsherrill, line 174: You might also add a comment above to make sure this gets updated too if that changes

jlsherrill:
tested and worked great, just one comment

jlsherrill (approved):

swadeley:
Hi

Change is now instant. Tested with GPRS setting (In Firefox, F12 -> Network, select "No thorttling", select GPRS).

PR #93, opened 2023-01-30
Fixes 1141: A few issues made apparent by long loading times

Andrewgdewar:
The table skeleton (seen when loading the main table) was overlapping the filter inputs.
The Add button on popular repositories was not being correctly disabled when clicked, allowing for multiple clicks and errors.


jlsherrill:
https://issues.redhat.com/browse/HMS-1141

jlsherrill:
One additional bug i notice, if you click multiple 'add' or 'remove' button in a row before the first is finished, the previous buttons you've clicked will re-enable and become clickable again.  Do you want to fix that as part of this? or a new bug?

jlsherrill:
also, if you select multiple items and do a bulk add, their add buttons do not become disabled

Andrewgdewar:
I should likely add a ticket to optimistically update these calls to make a much smoother experience. The current implementation can work for now though.

jlsherrill (approved):

PR #94, opened 2023-02-01
Fixes 579: add status filter

jlsherrill:


jlsherrill:
Todo: 

* [ ] Tests?

Andrewgdewar:
> Todo:
> 
> * [ ]  Tests?

Hmm yeah. Tests would be good but rather complicated 🙄.  
I probably should have added some earlier.

Andrewgdewar (approved):
ACK!

Andrewgdewar:
I'll add some tests to this later.

swadeley:
Hi, it is I, with a question. If I select Invalid, I only see Invalid repos. However, if I then select Valid as well, the Valid repo comes back but in the middle of the two Invalid repos, it does not sort them again. Is that to be expected? It only sorts once?

Before:
Screenshot 2023-02-02 at 10-19-07 contentSources console redhat com

After:

Screenshot 2023-02-02 at 10-19-52 contentSources console redhat com


Thank you.

jlsherrill:
@swadeley  i'm not sure i can quite reproduce what you mean, or maybe i don't fully understand what you mean.  In your screenshot, both shots seem sorted by name?

swadeley:
> @swadeley i'm not sure i can quite reproduce what you mean, or maybe i don't fully understand what you mean. In your screenshot, both shots seem sorted by name?

ahh, Filter by Status not equal to Sorted by Status, OK, sorry. So its working as designed then.

PR #95, opened 2023-02-14
Fixes 1173: Add optimistic updates to popular repositories

Andrewgdewar:
This change makes it so that the Add/Delete options in popular repositories update the list "optimistically", IE before getting a response from the API. 

One should now be able to perform back to back to actions when clicking add or remove (not both) .
Add
Deleting

FYI QE: To test this make sure to [slow down your network speed ](https://www.browserstack.com/guide/how-to-perform-network-throttling-in-chrome)in the browser.

### DIFF
--- a/src/services/Content/ContentQueries.ts
+++ b/src/services/Content/ContentQueries.ts
@@ -141,6 +141,72 @@ export const useAddContentQuery = (queryClient: QueryClient, request: CreateCont
   });
 };
 
+export const useAddPopularRepositoryQuery = (
+  queryClient: QueryClient,
+  request: CreateContentRequest,
+  page: number,
+  perPage: number,
+  filterData?: FilterData,
+) => {
+  const { notify } = useNotification();
+  const popularRepositoriesKeyArray = [
+    POPULAR_REPOSITORIES_LIST_KEY,
+    page,
+    perPage,
+    undefined,
+    ...Object.values(filterData || {}),
+  ];
+  const filteredRequest = request.filter((item) => !!item);
+  return useMutation(() => AddContentListItems(filteredRequest), {
+    onMutate: async () => {
+      const { name } = filteredRequest[0];
+      // Cancel any outgoing refetches (so they don't overwrite our optimistic update)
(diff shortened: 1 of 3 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-1173

jlsherrill:
functionally this seems great.   Looking at the code now 🔮

src/services/Content/ContentQueries.ts
```diff
@@ -195,6 +261,62 @@ export const useValidateContentList = () => {
...
+      const previousPopularData: Partial<PopularRepositoriesResponse> =
+        queryClient.getQueryData(popularRepositoriesKeyArray) || {};
+
+      queryClient.setQueryData(popularRepositoriesKeyArray, () => ({
+        ...previousPopularData,
+        data: previousPopularData.data?.map((data) => {
+          if (data.uuid === uuid) {
+            return { ...data, uuid: undefined };
+          }
+          return data;
```
jlsherrill, line 292: I'm probably not reading this correctly, but it looks like it optimistically updates the list to include the item but with a uuid of undefined?

swadeley:
Hi, I tested with throttle set to GPRS.

You can click rapidly on Remove and it all works as expected. For Add, you need to click a bit slower, you need to wait for the greyed, disabled, buttons above to turn red.

New action is much improved.

QE ACK pending question above about UUID undefined.

Thank you

src/services/Content/ContentQueries.ts
Andrewgdewar: This is accurate. I guess I could use "", as that make !uuid evaluate to true.

swadeley: @Andrewgdewar Re the above comment, are we waiting for one more change?

swadeley:
Hello

I tested again to make sure I have all the commits.
Throttled network speed to GPRS setting.

I confirm you can click as fast as you like to remove. On adding three repos in a row, it is possible to click so fast that a button does not respond. However, clicking it again works and it is unlikely to happen with average clicking speeds.

Thank you

jlsherrill (approved):

PR #104, opened 2023-03-22
Fixes 533: add introspect button and popover

rverdile:
This adds:
- "Introspect Now" action to the kebab
- Replaces the status tooltip with a popover that shows some new information and has a "retry" button for introspection

This depends on this backend PR: https://github.com/content-services/content-sources-backend/pull/221

Todo:
- [x] tests
- [x] merge backend PR

### DIFF
--- a/src/Pages/ContentListTable/components/StatusIcon.test.tsx
+++ b/src/Pages/ContentListTable/components/StatusIcon.test.tsx
@@ -1,27 +1,73 @@
-import { render } from '@testing-library/react';
+import { act, render, waitFor } from '@testing-library/react';
 import StatusIcon from './StatusIcon';
+import { ContentItem } from '../../../services/Content/ContentApi';
+
+const defaultData: ContentItem = {
+  uuid: '',
+  name: '',
+  package_count: 100,
+  url: '',
+  status: 'Pending',
+  account_id: '',
+  org_id: 'acme',
+  distribution_arch: '',
+  gpg_key: '',
+  distribution_versions: [''],
+  last_introspection_error: '',
+  last_introspection_time: '2023-03-07 17:13:48.619192 -0500 EST',
+  failed_introspections_count: 0,
+  metadata_verification: false,
+};
+
+jest.mock('../../../middleware/AppContext', () => ({
+  useAppContext: () => ({ rbac: { read: true, write: true } }),
+}));
 
 it('Render with Pending status', () => {
-  const { queryByText } = render(<StatusIcon status='Pending' />);
+  const { queryByText } = render(<StatusIcon rowData={{ ...defaultData, status: 'Pending' }} />);
 
   const SelectComponent = queryByText('In progress');
   expect(SelectComponent).toBeInTheDocument();
 });
 
 it('Render with Valid status', () => {
-  const { queryByText } = render(<StatusIcon status='Valid' />);
+  const { queryByText } = render(<StatusIcon rowData={{ ...defaultData, status: 'Valid' }} />);
   const SelectComponent = queryByText('Valid');
   expect(SelectComponent).toBeInTheDocument();
 });
 
 it('Render with Invalid status', () => {
-  const { queryByText } = render(<StatusIcon status='Invalid' />);
+  const { queryByText } = render(<StatusIcon rowData={{ ...defaultData, status: 'Invalid' }} />);
   const SelectComponent = queryByText('Invalid');
   expect(SelectComponent).toBeInTheDocument();
 });
 
-it('Render with Unavailable status', () => {
-  const { queryByText } = render(<StatusIcon status='Unavailable' />);
+it('Render with Unavailable status', async () => {
+  const { queryByText } = render(
+    <StatusIcon rowData={{ ...defaultData, status: 'Unavailable' }} />,
+  );
   const SelectComponent = queryByText('Unavailable');
   expect(SelectComponent).toBeInTheDocument();
+
+  await act(async () => {
+    SelectComponent?.click();
+  });
+
+  await waitFor(() => {
+    expect(queryByText('Retry')).toBeInTheDocument();
+  });
+});
+
+it('Render with Invalid status', async () => {
+  const { queryByText } = render(<StatusIcon rowData={{ ...defaultData, status: 'Invalid' }} />);
+  const SelectComponent = queryByText('Invalid');
+  expect(SelectComponent).toBeInTheDocument();
+
+  await act(async () => {
+    SelectComponent?.click();
+  });
+
+  await waitFor(() => {
+    expect(queryByText('Retry')).toBeInTheDocument();
+  });
 });
--- a/src/Pages/ContentListTable/components/StatusIcon.tsx
+++ b/src/Pages/ContentListTable/components/StatusIcon.tsx
@@ -3,7 +3,16 @@ import {
   ExclamationCircleIcon,
   ExclamationTriangleIcon,
 } from '@patternfly/react-icons';
-import { Flex, FlexItem, Spinner, Tooltip } from '@patternfly/react-core';
+import {
+  Button,
+  Flex,
+  FlexItem,
+  Popover,
+  Spinner,
+  Stack,
+  StackItem,
+  Tooltip,
+} from '@patternfly/react-core';
 import StatusText from './StatusText';
 import {
   global_danger_color_100,
@@ -11,6 +20,12 @@ import {
   global_warning_color_100,
 } from '@patternfly/react-tokens';
 import { createUseStyles } from 'react-jss';
+import dayjs from 'dayjs';
+import relativeTime = require('dayjs/plugin/relativeTime');
+import { ContentItem } from '../../../services/Content/ContentApi';
+import ConditionalTooltip from '../../../components/ConditionalTooltip/ConditionalTooltip';
+import { useAppContext } from '../../../middleware/AppContext';
+dayjs.extend(relativeTime);
 
 const red = global_danger_color_100.value;
 const green = global_success_color_100.value;
@@ -22,13 +37,71 @@ const useStyles = createUseStyles({
     margin: '-3px 0',
   },
 });
+
 interface Props {
-  status?: string;
-  error?: string;
+  rowData: ContentItem;
+  retryHandler?: (string) => Promise<void>;
 }
 
-const StatusIcon = ({ status, error }: Props) => {
+interface DescriptionProps {
+  error?: string;
+  count?: number;
+  time?: string;
+}
+
+interface FooterProps {
+  retryHandler?: (string) => Promise<void>;
+  uuid: string;
+}
+
+const PopoverDescription = ({ error, count, time }: DescriptionProps) => {
+  const timeText = time === '' || time === undefined ? 'Never' : dayjs(time).fromNow();
+  return (
+    <div>
+      {error}
+      <Flex>
+        <Stack className='pf-u-mt-sm'>
+          <StackItem className='pf-u-font-weight-bold'>Last introspection</StackItem>
+          <StackItem> {timeText} </StackItem>
+        </Stack>
+        <Stack className='pf-u-mt-sm'>
+          <StackItem className='pf-u-font-weight-bold'>Failed attempts</StackItem>
+          <StackItem> {count} </StackItem>
+        </Stack>
+      </Flex>
+    </div>
+  );
+};
+
+const PopoverFooter = ({ retryHandler, uuid }: FooterProps) => {
+  const { rbac } = useAppContext();
+  return (
+    <Flex>
+      <ConditionalTooltip
+        content='You do not have the required permissions to perform this action.'
+        show={!rbac?.write}
+        setDisabled
+      >
+        <Button variant='link' isInline onClick={() => (retryHandler ? retryHandler(uuid) : null)}>
+          Retry
+        </Button>
+      </ConditionalTooltip>
+    </Flex>
+  );
+};
+
+const StatusIcon = ({
+  rowData: {
+    uuid,
+    status,
+    failed_introspections_count: failedIntrospectionsCount,
+    last_introspection_time: lastIntrospectionTime,
+    last_introspection_error: error,
+  },
+  retryHandler,
+}: Props) => {
   const classes = useStyles();
+
   switch (status) {
     case 'Valid':
       return (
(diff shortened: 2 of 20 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-533

swadeley:
Hi

test run failed:

`15:38:17   console.error
15:38:17     Error: Uncaught [TypeError: (0 , ContentQueries_1.useIntrospectRepositoryMutate) is not a function]`

thank you

src/Pages/ContentListTable/components/StatusPopover.tsx
rverdile, line 1: I originally made this a component then I felt like a component wasn't necessary. I'll delete this unless using a component is preferred.

src/services/Content/ContentQueries.ts
```diff
@@ -443,3 +446,85 @@ export const useGetPackagesQuery = (
...
+    const previousData: Partial<ContentListResponse> =
+      queryClient.getQueryData(contentListKeyArray) || {};
+
+      const previousStatus = previousData.data?.find((data) => item.uuid == data.uuid)?.status
+      const newData = cloneDeep(previousData)
+      if (previousStatus) {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        newData.data.filter((data) => item.uuid == data.uuid).at(0).status = 'Retrying'
+      }
```
rverdile, line 480: is there a cleaner way to do this?

src/Pages/ContentListTable/components/StatusIcon.test.tsx
```diff
@@ -1,27 +1,53 @@
...
 });
 
 it('Render with Unavailable status', () => {
-  const { queryByText } = render(<StatusIcon status='Unavailable' />);
+  const { queryByText } = render(<StatusIcon rowData={{ ...defaultData, status: 'Unavailable' }} />);
   const SelectComponent = queryByText('Unavailable');
   expect(SelectComponent).toBeInTheDocument();
+  act(() => {SelectComponent?.dispatchEvent(new MouseEvent('click'))});
+  expect(queryByText('Retry')).toBeInTheDocument()
+});
```
rverdile, line 47: having trouble getting the popover to pop-up in tests

src/services/Content/ContentQueries.ts
Andrewgdewar: Yes an no, I think there are alternatives, that you could use with variables. 
Instead of filter for example you could use findIndex > store that in a variable > and target the specific value in newData.data[foundIndex].status ="stuff".

Also, anytime you use a variable chain on a filter, you should likely safe those values with ?. 
And then make the assignment conditional.

Andrewgdewar:
Screen Shot 2023-03-30 at 3 31 32 PM
Looks like you broke default value loading for the status page. 
You likely want to make sure that your default returned values are being managed correctly to make the loading items appear in the status problem. 
Also in the case of the something having never been introspected, the last introspected value (at least in the front-end) should likely say "Never"

src/Pages/ContentListTable/components/StatusIcon.tsx
```diff
@@ -43,31 +71,50 @@ const StatusIcon = ({ status, error }: Props) => {
...
+                position='left'
+                footerContent={popoverFooter(retryHandler, uuid)}
+              >
+                <StatusText color='gold' isLink>Unavailable</StatusText>
+              </Popover>
             </FlexItem>
           </Flex>
-        </Tooltip>
       );
     case 'Pending':
```
Andrewgdewar, line 116: case '':

src/Pages/ContentListTable/components/StatusPopover.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+interface Props {
+  status: 'Invalid' | 'Unavailable';
+  error?: string;
+  lastIntrospectionTime?: string;
+  failedIntrospectionsCount?: number
+  children?: React.ReactElement;
+}
+
+const popoverDescription = (error: string | undefined, count: number | undefined, time: string | undefined) =>
```
Andrewgdewar, line 13: (error?: string , count?: number , time?: string)

src/services/Content/ContentQueries.ts
Andrewgdewar: Why was "Retrying" necessary here. I'd expect Pending to work just as well.

src/Pages/ContentListTable/components/StatusPopover.tsx
```diff
@@ -0,0 +1,65 @@
...
+    </Stack>
+    <Stack className="pf-u-mt-sm">
+      <StackItem className="pf-u-font-weight-bold">Failed attempts</StackItem>
+      <StackItem> {count} </StackItem>
+    </Stack>
+  </div>
+
+const popoverFooter = () => <Button variant="link" isInline>Retry</Button>
+
+const StatusPopover = ({ status, error, lastIntrospectionTime, failedIntrospectionsCount, children }: Props) => {
```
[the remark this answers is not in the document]
rverdile: Nope. I mentioned this in my first comment, but I originally made this a component. Then I felt like a component wasn't necessary, so I moved to putting things in `StatusIcon.tsx`. I left this here in case a component was preferred.

The things that stay consistent between the two different popovers are the description and the footer, which can just use a shared function. The rest of the component was just a switch statement, which already exists in `StatusIcon.tsx`. So, I felt like that was simpler to just have two shared functions than sharing a whole new component.

src/services/Content/ContentQueries.ts
rverdile: I made the Retrying status because Pending implied that no packages were available (since the repository had never been introspected). With Retrying, packages could be available already. And you can see that in the UI if you have a repository and "retry", the package count doesn't switch to "N/A".

Andrewgdewar: I would still disable the package count while introspecting in any case, as that data may change. UNLESS, when your retrying and successful you somehow invalidate that package list cache.

rverdile: I made it N/A for retrying. It may have been necessary for me to add a new status...I'm trying to think if there's another reason for an entirely different status. If it not, it probably makes more sense to just think about the Pending status differently.

rverdile: And in response to "is there a cleaner way to do this". I can't really figure out another way to do it where typescript doesn't also complain. But if you have a specific way of doing it that you like that you like more, feel free to show me how :).

src/Pages/ContentListTable/components/StatusIcon.test.tsx
rverdile: WaitFor worked!

src/Pages/ContentListTable/components/StatusIcon.tsx
Andrewgdewar: This shouldn't be needed if we fix [this](https://issues.redhat.com/browse/HMS-1564)

src/services/Content/ContentQueries.ts
Andrewgdewar: I'll tinker on monday when I do another review.

rverdile:
Removed "Retrying" and replaced with "Pending"

src/Pages/ContentListTable/components/StatusIcon.tsx
```diff
@@ -43,31 +77,50 @@ const StatusIcon = ({ status, error }: Props) => {
...
             <FlexItem>
-              <StatusText color='red'>Invalid</StatusText>
+              <Popover
+                aria-label="invalid popover"
+                alertSeverityVariant='danger'
+                headerContent="Invalid"
+                headerIcon={<ExclamationCircleIcon/>}
+                bodyContent={popoverDescription(error, failedIntrospectionsCount, lastIntrospectionTime)}
+                position='left'
+                footerContent={popoverFooter(retryHandler, uuid)}
```
Andrewgdewar, line 92: Make this a component, a function type return will force un-needed rerenders.
```suggestion
                footerContent={<PopoverComponent  props=whatevs >}
```

In this case though, with such small

Andrewgdewar: Same with description above.

src/Pages/ContentListTable/components/StatusIcon.tsx
```diff
@@ -22,13 +26,43 @@ const useStyles = createUseStyles({
...
-  status?: string;
-  error?: string;
+  rowData: ContentItem;
+  retryHandler?: (string) => Promise<void>;
 }
 
-const StatusIcon = ({ status, error }: Props) => {
+const popoverDescription = (error?: string, count?: number, time?: string) => {
+  if (time === '') {
+    time = 'Never'
```
Andrewgdewar, line 37: In typescript/javscript in general, we try to avoid re-assigning props/imported variables.

src/Pages/ContentListTable/components/StatusIcon.tsx
```diff
@@ -22,13 +26,43 @@ const useStyles = createUseStyles({
...
+      <StackItem className="pf-u-font-weight-bold">Failed attempts</StackItem>
+      <StackItem> {count} </StackItem>
+    </Stack>
+  </div>
+  )
+}
+
+const popoverFooter = (retryHandler: any, uuid: string) => <Button variant="link" isInline onClick={() => retryHandler(uuid)}>Retry</Button> // TODO type retryHandler correctly
+
+const StatusIcon = ({ rowData, retryHandler }: Props) => {
```
Andrewgdewar, line 56:
```suggestion
const StatusIcon = ({ rowData:{
  uuid,
  status,
  failed_introspections_count: failedIntrospectionsCount,
  last_introspection_time: lastIntrospectionTime,
  last_introspection_error: error,
  }, retryHandler }: Props) => {
```

src/Pages/ContentListTable/components/StatusIcon.tsx
```diff
@@ -22,13 +26,43 @@ const useStyles = createUseStyles({
...
   const classes = useStyles();
+  const {
+  uuid: uuid,
+  status: status,
+  failed_introspections_count: failedIntrospectionsCount,
+  last_introspection_time: lastIntrospectionTime,
+  last_introspection_error: error,
+  } = rowData
```
Andrewgdewar, line 64: Simplify by plopping this in the props above if not using rowData directly.

src/Pages/ContentListTable/components/StatusIcon.tsx
```diff
@@ -43,31 +77,50 @@ const StatusIcon = ({ status, error }: Props) => {
...
+                footerContent={popoverFooter(retryHandler, uuid)}
+              >
+                <StatusText color='gold' isLink>Unavailable</StatusText>
+              </Popover>
             </FlexItem>
           </Flex>
-        </Tooltip>
       );
     case 'Pending':
+    case '':
```
Andrewgdewar, line 123: After thinking about this, it may make more sense not to add this here. 
But rather add it to the check for the polling script within react-query. 
As say if we have a "" in the future with very few items and no "Pending" statuses. The update script would never run. 

That way we aren't masking the mistake from the API (or expecting it) and we are still refreshing the page every 10 seconds with the expectation that backend should updated the value.

src/Pages/ContentListTable/components/StatusText.tsx
```diff
@@ -20,13 +20,21 @@ const useStyles = createUseStyles({
 
 interface Props {
   color: 'red' | 'green' | 'gold' | 'blue';
+  isLink?: boolean;
```
Andrewgdewar, line 23: AH cool, so here is a good time to introduce the extend type for createUseStyles. 
With it you could do the following 

```
import { createUseStyles } from 'react-jss';
import {
  global_danger_color_200,
  global_info_color_200,
  global_success_color_200,
  global_warning_color_200,
} from '@patternfly/react-tokens';

const red = global_danger_color_200.value;
const green = global_success_color_200.value;
const gold = global_warning_color_200.value;
const blue = global_info_color_200.value;

interface Props {
  color: 'red' | 'green' | 'gold' | 'blue';
  isLink?: boolean;
  children?: React.ReactNode;
}

const useStyles = ({ isLink }: Props) =>
  createUseStyles({
    fontStyle: {
      fontWeight: 'bold',
      fontSize: '14px',
      ...(isLink
        ? {
            textDecoration: 'grey dotted underline',
            cursor: 'pointer',
          }
        : {}),
    },
    red: { extend: 'fontStyle', color: red },
    green: { extend: 'fontStyle', color: green },
    gold: { extend: 'fontStyle', color: gold },
    blue: { extend: 'fontStyle', color: blue },
  });

const StatusText = (props: Props) => {
  const classes = useStyles(props)();
  const { color, children } = props;

  return <span className={classes[color]}>{children}</span>;
};

export default StatusText;

```

src/Pages/ContentListTable/components/StatusIcon.test.tsx
```diff
@@ -1,27 +1,67 @@
...
+it('Render with Unavailable status', async () => {
+  const { queryByText } = render(<StatusIcon rowData={{ ...defaultData, status: 'Unavailable' }} />);
   const SelectComponent = queryByText('Unavailable');
   expect(SelectComponent).toBeInTheDocument();
+
+  await act(async () => {
+    SelectComponent?.click();
+  });
+
+  await waitFor(() => {
```
Andrewgdewar, line 50: Yup!

Andrewgdewar:
Make sure to run "npm run format" before pushing up changes if not running a formatter.

src/Pages/ContentListTable/components/StatusIcon.tsx
rverdile: I did the first thing

src/Pages/ContentListTable/components/StatusText.tsx
Andrewgdewar: Oh a note here, I didn't actually test the above code, it's just an example of what I imagine it to look like.

rverdile: ah, well I think it worked

Andrewgdewar:
This is looking good, I highlighted one point around permissions, once that is resolved this gets my ACK.

Andrewgdewar (approved):
And my ACK... 
One note, the backend API should have accompanying code to limit the retry action to be in parity with the front-end permissions prior to merging.

swadeley:
Hi, backed PR was merged four days ago. I see build test failed "SyntaxError: Unexpected token 'export'" about five days ago.
I'll trigger a new test.

src/Pages/ContentListTable/components/StatusIcon.tsx
```diff
@@ -22,13 +37,71 @@ const useStyles = createUseStyles({
...
+          <StackItem className='pf-u-font-weight-bold'>Failed attempts</StackItem>
+          <StackItem> {count} </StackItem>
+        </Stack>
+      </Flex>
+    </div>
+  );
+};
+
+const PopoverFooter = ({ retryHandler, uuid }: FooterProps) => {
+  const { rbac } = useAppContext();
```
Andrewgdewar, line 77: This is what is making your test fail (I think). You'll want to mock the useAppContext to return this value.

swadeley:
Hi @rverdile 

The new "Introspect Now" action works  thank you. However I do not see "tooltip with a popover that shows some new information and has a "retry" button". During introspection, there is a tooltip that says "Repository is being introspected", but after that completes there is no more tooltip for Status.

Thank you

rverdile:
> seems there is some change in the table when applying a filter.

I'm not sure what's happening here. When I play around with the PR manually I don't see anything wrong. I did add a new filter for the column "Last Introspection". Could that have something to do with it?

swadeley:
Hi @rverdile 
No, sorry, I was not testing with repo that failed introspection. I will test again.

Our test is failing to open the Edit dialogue. So far I cannot spot what has changed.

Thank you

swadeley:
Hi @rverdile 

I confirm with bad  repo the new tooltip it works

```
danger alert:Invalid
Cannot fetch https://fixtures.pulpproject.org/404.html/repodata/repomd.xml: 404
Last introspection
a few seconds ago
Failed attempts
1 
```

There is a Retry button on bottom left

Success alert:Repository introspection in progress

that works


Thank you

PR #125, opened 2023-06-27
Fixes 1661: Fix debounce logic and errors

Andrewgdewar:
## Summary

Fixes an issue where the gpgKey in Add content modal would flash a success check before finishing validation. 

Added loading logic to Save button in AddContent

Both addContent/editContent loading button logic is dependent on the same variables as the modal "OUIA_safe", this should ensure tests work as expected when using both of these.

Fixes an issue with css around the gpgKey success/error badge being partially obscured

Fixes an issue with gpgkey (and other variable) debounce logic causing unneeded api calls to validate.

Other changes: 
- Moved custom hooks to their own folder


swadeley (approved):
Excellent

PR #127, opened 2023-07-12
Fixes 2063: Add bulk delete

dpang314:
## Summary

Adds check boxes on the left in the content list table that allows for selecting multiple repositories. Repositories can be selected across pages. To the right of the "Add Repositories" button there is a new kebab button. With a button to delete all checked repositories.

Also adds the same kebab on the popular repositories page.

Currently, when an error occurs in deletion, only the first error is displayed.

One potential issue that occurs is whenever a modal opens for the first time on a page (such as add repository or edit repository) with the new routing system, clicking cancel causes the page to refresh, clearing all the checked repositories even though no action has occurred.

### When rbac.write permission is true:

On `/insights/content`
- When no repositories are checked, the kebab to the right of the "Add repositories" button should be enabled. Clicking on the kebab reveals a button that reads "Remove selected repositories" which is disabled and on hover, shows a popup reading "Make a selection below to add multiple repositories"
- Repositories with a "Pending" status should not be able to be selected
- Ticking the select all checkbox selects all repositories that aren't pending
- Deleting repositories will change the page if the deletion causes the page to no longer exist (e.g. deleting all the repositories on the last page)
- Deleting over 100 repositories displays an error
- The "select all" checkbox column header is disabled when there are no checkable repositories on the current page (i.e. they are all pending)
- Checked repositories should be reset when repositories are added, edited, deleted, and introspected
- Checked repositories are not reset when an error occurs in any of those actions
- Repositories can be checked across pages

On `insights/content/popular-repositories`
- Both repositories that have and have not been added can be selected at the same time
- When no repositories are checked, the button next to the filter should read "Add selected repositories." It should be disabled and on hover, show a popup reading "Make a selection below to add multiple repositories" 
- When all checked repositories have already been added, the behavior should be the same
- When no repositories are checked, the kebab next to the button should be enabled. Clicking on the kebab reveals a button that reads "Remove selected repositories" which is disabled and on hover, shows a popup reading "Make a selection below to add multiple repositories"
- When all checked repositories have not been added, the behavior should be the same
- When there is at least 1 repository checked that has not been added, the add button should read "Add # repositories"
- When there is at least 1 repository checked that has been added, the delete button should read "Remove # repositories"
- Both of the above conditions should work at the same time
- When there are both added and not added checked repositories, clicking the add button should only add the repositories that have not been added, and clicking the delete button should only delete the repositories that have been added
- Repositories can be checked across pages. This can be tested by filtering to "EPEL 9", checking it, filtering to "EPEL 8", checking it, and then clearing the filter. Both repositories should still be checked
- Checked repositories should be reset when repositories are added individually, deleted individually, bulk added, and bulk deleted
- Checked repositories are not reset when an error occurs in any of those actions
- Adding over 20 repositories at once returns an error
- Deleting over 100 repositories at once returns an error

### When rbac.write permission is false:

On `/insights/content`
- "Add repositories" button is disabled with "no permission" popup on hover
- Kebab to the right of that button is disabled with same popup on hover
- No checkboxes appear on the left column of the table

On `insights/content/popular-repositories`
- "Add selected repositories" button is disabled with "no permission" popup on hover
- Kebab to the right of that button is disabled with same popup on hover
- No checkboxes appear on the left column of the table

### DIFF
--- a/src/Pages/ContentListTable/components/ContentListFilters.tsx
+++ b/src/Pages/ContentListTable/components/ContentListFilters.tsx
@@ -28,6 +28,7 @@ interface Props {
   setFilterData: (filterData: FilterData) => void;
   filterData: FilterData;
   atLeastOneRepoChecked: boolean;
+  numberOfReposChecked: number;
   deleteCheckedRepos: () => void;
 }
 
@@ -65,6 +66,7 @@ const ContentListFilters = ({
   setFilterData,
   filterData,
   atLeastOneRepoChecked,
+  numberOfReposChecked,
   deleteCheckedRepos,
 }: Props) => {
   const classes = useStyles();
@@ -279,6 +281,7 @@ const ContentListFilters = ({
         </ConditionalTooltip>
         <ContentActions
           atLeastOneRepoChecked={atLeastOneRepoChecked}
+          numberOfReposChecked={numberOfReposChecked}
           deleteCheckedRepos={deleteCheckedRepos}
         />
       </FlexItem>
--- a/src/Pages/ContentListTable/components/ContentListFilters.tsx
+++ b/src/Pages/ContentListTable/components/ContentListFilters.tsx
@@ -50,6 +50,11 @@ const useStyles = createUseStyles({
     left: '-5px',
     pointerEvents: 'none',
   },
+  // Needed to fix styling when "Add repositories" button is disabled
+  repositoryActions: {
+    display: 'flex',
+    flexDirection: 'row',
+  },
 });
 
 const statusValues = ['Invalid', 'Pending', 'Unavailable', 'Valid'];
@@ -256,7 +261,7 @@ const ContentListFilters = ({
           <FlexItem>{Filter}</FlexItem>
         </InputGroup>
       </FlexItem>
-      <FlexItem>
+      <FlexItem className={classes.repositoryActions}>
         <ConditionalTooltip
           content='You do not have the required permissions to perform this action.'
           show={!rbac?.write}
@@ -272,16 +277,10 @@ const ContentListFilters = ({
             Add repositories
           </Button>
         </ConditionalTooltip>
-        <ConditionalTooltip
-          content='You do not have the required permissions to perform this action.'
-          show={!rbac?.write}
-          setDisabled
-        >
-          <ContentActions
-            atLeastOneRepoChecked={atLeastOneRepoChecked}
-            deleteCheckedRepos={deleteCheckedRepos}
-          />
-        </ConditionalTooltip>
+        <ContentActions
+          atLeastOneRepoChecked={atLeastOneRepoChecked}
+          deleteCheckedRepos={deleteCheckedRepos}
+        />
       </FlexItem>
       <Hide
         hide={
--- a/src/Pages/ContentListTable/components/ContentListFilters.tsx
+++ b/src/Pages/ContentListTable/components/ContentListFilters.tsx
@@ -22,6 +22,7 @@ import { isEmpty } from 'lodash';
 import { useAppContext } from '../../../middleware/AppContext';
 import ConditionalTooltip from '../../../components/ConditionalTooltip/ConditionalTooltip';
 import { useNavigate } from 'react-router-dom';
+import DeleteKebab from '../../../components/DeleteKebab/DeleteKebab';
 
 interface Props {
   isLoading?: boolean;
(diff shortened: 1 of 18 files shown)


src/Pages/ContentListTable/components/ContentListFilters.tsx
```diff
@@ -238,14 +252,18 @@ const ContentListFilters = ({ isLoading, setFilterData, filterData }: Props) =>
           <FlexItem>{getSelectionByType()}</FlexItem>
         </InputGroup>
       </FlexItem>
-      <FlexItem>
+      <FlexItem className={classes.repositoryActions}>
```
Andrewgdewar, line 255: [here](https://www.patternfly.org/v4/layouts/flex#flexitem)

swadeley:
Hi

Added two popular repos. New keybab option shows `Remove 0 repositories`

Selected all repos (there are two repos at this stage), now it shows `Remove 2 repositories`

image

swadeley:
Hi

- When no repositories are checked, the "Remove all" button in the kebab should be disabled
      Confirmed (see above)
  
- Repositories with a "Pending" status should not be able to be selected
      Confirmed.
- Ticking the select all checkbox selects all repositories that aren't pending
      Confirmed.
- Deleting repositories will change the page if the deletion causes the page to no longer exist (e.g. deleting all the repositories on the last page)
   Confirmed

swadeley:
Hi

Re:

- Using the single delete (kebab button in each row) on a checked repository removes that repository from the checked repositories (doesn't cause any errors when bulk deleting later)
              I found that clears the selection of the other repos.

Andrewgdewar:
> Hi
> 
> Re:
> 
> * Using the single delete (kebab button in each row) on a checked repository removes that repository from the checked repositories (doesn't cause any errors when bulk deleting later)
>   I found that clears the selection of the other repos.

Anytime a C(r)UD action is taken, all selections should be removed. This was requested by myself. This is due to the List being refreshed/pulled when these types of actions take place.

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.tsx
```diff
@@ -134,44 +139,121 @@ const PopularRepositoriesTable = () => {
     { searchQuery: debouncedSearchValue } as FilterData,
   );
 
+  const repoToRequestItem = (repo: PopularRepository): CreateContentRequestItem => {
```
Andrewgdewar, line 142: You should take any functions like this (that don't have internal imported values) and plop them in a "helper.ts"

src/services/Content/ContentQueries.ts
```diff
@@ -204,7 +207,7 @@ export const useAddPopularRepositoryQuery = (
         };
         queryClient.setQueryData(popularRepositoriesKeyArray, previousData);
       }
-      errorNotifier('Error deleting item from popularRepo', 'An error occurred', err);
+      errorNotifier('Error adding item from popularRepo', 'An error occurred', err);
```
Andrewgdewar, line 210: Thanks!

swadeley:
Hi

Re "Repositories can be checked across pages"
I have 15 repos, I changed pagination to 10 per page.
I select all, it only selects the ten on the page. Is that expected?

Andrewgdewar:
> Hi
> 
> Re "Repositories can be checked across pages" I have 15 repos, I changed pagination to 10 per page. I select all, it only selects the ten on the page. Is that expected?

That is expected. As we don't have the data to select the rest.

Andrewgdewar:
Screen Shot 2023-07-31 at 11 03 23 AM
@swadeley Question (not being trite) should we have a PR Build check if it's always broken? I don't think I've seen a passing check in the last 6 months for the front-end.

Andrewgdewar (approved):
and my ack!

swadeley:
> Screen Shot 2023-07-31 at 11 03 23 AM
> 
> @swadeley Question (not being trite) should we have a PR Build check if it's always broken? I don't think I've seen a passing check in the last 6 months for the front-end.

If a PR build is broken it should be investigated. Next time you see one, ask yourself, does my PR break something?  If the answer is no or unsure, ask in chat.

swadeley:
Hi

RBAC testing

When rbac.write permission is false:

On /insights/content

    * "Add repositories" button is disabled with "no permission" popup on hover
           Confirmed, I see "You do not have the required permissions to perform this action."
    * Kebab to the right of that button is disabled with same popup on hover
          Confirmed, I see "You do not have the required permissions to perform this action."
     * No checkboxes appear on the left column of the table
           Confirmed

On insights/content/popular-repositories

    * "Add selected repositories" button is disabled with "no permission" popup on hover
           Confirmed, I see "You do not have the required permissions to perform this action."
    * Kebab to the right of that button is disabled with same popup on hover
           Confirmed, I see "You do not have the required permissions to perform this action."
    *  No checkboxes appear on the left column of the table
           Confirmed

PR #131, opened 2023-08-03
Fixes 2271: Make url derived from history object for nav

Andrewgdewar:
## Summary
This was causing an issue where root urls of non "insights" would reroute to "settings". 

Expected behaviour is now to inherit the root url, cross domain jumping should not occur.

## Testing steps

- Open and close all modals when on the "insights" root address. 
- On close of modal the insights address should remain, and not be replaced by "settings"

### DIFF
--- a/src/Hooks/useRootPath.tsx
+++ b/src/Hooks/useRootPath.tsx
@@ -1,0 +1,14 @@
+import { useMemo } from 'react';
+import { useLocation } from 'react-router-dom';
+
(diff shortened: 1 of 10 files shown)


jlsherrill (approved):
ACK

swadeley:
Hi

tested OK with Add Repo modal, Edit modal, and Packages modal (all of which exhibited the fault described in current version of stage).

Thank you

PR #143, opened 2023-09-12
Fixes 1700: Update edit/add modal checkmark functionality

Andrewgdewar:
## Summary

This updates the add/edit modals with the following functionality around validation: 

For the add modal: 
- The controller will focus on url errors first, even if the gpg key has the wrong value as it is dependant on the URL for validation.
- Only mandatory fields have a green check mark and the input gpg key box will only have a red mark if we enter the wrong value.
- if the name and gpgkey both have errors they will both show a red check as they are not hierarchally dependent.

For the Edit modal: 

- Validation will only show green checkmarks (showing updated values) if there are no errors showing, and when the validation API returns clear. 
- If an error is present, no checkmarks will show, forcing the user to focus on the error before confirming any other changes. 
- The same error flow (gpg key dependent on URL) is present from add modal, on the edit modal.

## Testing steps

- Step through the add/create modals, noting the above new behaviour. 
- Highlight any inconsistencies if found.

### DIFF
--- a/src/Pages/ContentListTable/components/AddContent/helpers.test.ts
+++ b/src/Pages/ContentListTable/components/AddContent/helpers.test.ts
@@ -1,5 +1,7 @@
+import { FormikErrors } from 'formik';
 import { ValidationResponse } from '../../../../services/Content/ContentApi';
 import {
+  FormikValues,
   REGEX_URL,
   failedFileUpload,
   isValidURL,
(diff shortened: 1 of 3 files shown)


jlsherrill:
* When i add or edit a repo that doesn't support metadata signatures, the option to select 'Package and metadata verification' is still present

Otherwise i think this is good

PR #144, opened 2023-09-12
Fixes 2312: Admin tasks interface append results

Andrewgdewar:
## Summary

- This is an attempted fix for the duplicate item error being appended in stage. 
- I was unable to replicated the issue locally, but have corrected a few issues in the code around paginating that could be causing this issue. 

## Testing steps

- One can check in ephemeral, but this likely needs to be pushed to stage to have its' functionality confirmed.


jlsherrill:
https://issues.redhat.com/browse/HMS-2312

jlsherrill (approved):

PR #155, opened 2023-10-23
Fixes 2190: added snapshot info to content list table

xbhouse:
## Summary

- Added "Snapshot" related info to the repositories list table (sketch reference [here](https://www.sketch.com/s/b55e75a8-9a53-4eb0-989e-aad208521aba/a/VEkGYmq))
- Displays "No snapshot yet" if there are no snapshots, otherwise displays "Last snapshot (creation time of last snapshot) Changes: " with # of additions and # of removals using the ChangedArrows component
- Displays nothing if feature not enabled
- Updated tests

## Testing steps

- Enable snapshot feature, snapshot info should be displayed
- When feature is disabled, it should look as before
- Depends on [this PR](https://github.com/content-services/content-sources-backend/pull/457)

### DIFF
--- a/src/Pages/ContentListTable/ContentListTable.tsx
+++ b/src/Pages/ContentListTable/ContentListTable.tsx
@@ -20,7 +20,7 @@ import {
   ThProps,
   Tr,
 } from '@patternfly/react-table';
-import { global_BackgroundColor_100 } from '@patternfly/react-tokens';
+import { global_BackgroundColor_100, global_Color_400 } from '@patternfly/react-tokens';
 import { useCallback, useMemo, useState } from 'react';
 import { createUseStyles } from 'react-jss';
 import {
@@ -76,15 +76,13 @@ const useStyles = createUseStyles({
     minWidth: '45px!important',
   },
   snapshotInfoText: {
-    color: 'grey',
+    color: global_Color_400.value,
     paddingRight: '20px',
   },
   inline: {
-    display: 'flex'
+    display: 'flex',
+    gap: '10px',
   },
-  snapshotInfoPadding: {
-    paddingTop: '10px'
-  }
 });
 
 const perPageKey = 'contentListPerPage';
@@ -461,19 +459,18 @@ const ContentListTable = () => {
                             }}
                           />
                         </Hide>
-                        <Td>
+                        <Td className={classes.inline}>
                           {name}
-                          <br />
                           <UrlWithExternalIcon href={url} />
-                          {features?.snapshots?.accessible && 
-                            <Flex className={classes.snapshotInfoPadding}>
+                          <Hide hide={!features?.snapshots?.accessible}>
+                            <Flex>
                               <FlexItem className={classes.snapshotInfoText}>
                                 {last_snapshot ? 
                                   `Last snapshot ${dayjs(last_snapshot?.created_at).fromNow()}` : 
                                   'No snapshot yet'
                                 }
                               </FlexItem>
-                              {last_snapshot &&
+                              <Hide hide={!last_snapshot}>
                                 <FlexItem className={classes.inline}>
                                   <FlexItem className={classes.snapshotInfoText}>Changes:</FlexItem>
                                   <ChangedArrows
@@ -483,9 +480,9 @@ const ContentListTable = () => {
                                                   (last_snapshot?.removed_counts?.['rpm.package'] || 0)}
                                   />
                                 </FlexItem>
-                              }
+                              </Hide>
                             </Flex>
-                          }
+                          </Hide> 
                         </Td>
                         <Td>{archesDisplay(distribution_arch)}</Td>
                         <Td>{versionDisplay(distribution_versions)}</Td>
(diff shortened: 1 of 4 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-2190

src/Pages/ContentListTable/ContentListTable.tsx
```diff
@@ -74,6 +75,16 @@ const useStyles = createUseStyles({
   checkboxMinWidth: {
     minWidth: '45px!important',
   },
+  snapshotInfoText: {
+    color: 'grey',
```
Andrewgdewar, line 79: So there are colors specific to this from patternfly if you look at the Header.tsx you'll see a similar value (likely to this color) referenced called: 
import { global_Color_100 } from '@patternfly/react-tokens';

global_Color_100.value

src/Pages/ContentListTable/ContentListTable.tsx
```diff
@@ -453,6 +465,27 @@ const ContentListTable = () => {
                           {name}
                           <br />
                           <UrlWithExternalIcon href={url} />
+                          {features?.snapshots?.accessible &&
```
Andrewgdewar, line 468: So this is a funny one, there are times when doing things like this will return the text value of "unknown" "undefined" or "false" into the final html.

So in this case using a ternary operator with an empty fragment would be ideal. 
BUT we do have a custom wrapper component that we can use instead that handles this, simply called
` <Hide hide={boolean}>`

You should likely use that here, and below where the "last_snapshot" is being checked.

src/Pages/ContentListTable/ContentListTable.tsx
```diff
@@ -453,6 +465,27 @@ const ContentListTable = () => {
                           {name}
                           <br />
```
Andrewgdewar, line 466: oh hrmm a `<br />` 

We only want to do this within spans/paragraph elements, I suppose I must have broken my own roles, oh dear.  

I good method to fix this issue and the spacing between each element here would be to make this Td us flex (or wrap the contents as you've done below). And then use the "gap" css feature to ensure equal distance in all children. 
https://css-tricks.com/almanac/properties/g/gap/

src/testingHelpers.tsx
```diff
@@ -208,3 +208,22 @@ export const defaultSnapshotItem: SnapshotItem = {
   },
   removed_counts: {},
 };
+
+export const defaultContentItemWithSnapshot: ContentItem = {
```
Andrewgdewar, line 212: Thanks for yoinking this out to this file! 👍

src/Pages/ContentListTable/ContentListTable.tsx
xbhouse: cool, changed to global_Color_400

src/Pages/ContentListTable/ContentListTable.tsx
xbhouse: ah gotcha! updated to use Hide

src/Pages/ContentListTable/ContentListTable.tsx
xbhouse: removed the br and made the changes you suggested, let me know if that's what you meant. the items in the Td seem to be spaced equally now

Andrewgdewar (approved):
Ack!

swadeley:
Hi, I created a repo with out snap shots enabled and then one with snapshots enabled:

image

swadeley:
Hi, I disabled snapshot support and on creating another repo with snapshot=false I still see the text "No snapshot yet" which I thought we should not see.

image

xbhouse:
thank you @swadeley. just pushed a fix, changed the context to check for features.snapshots.enabled instead of features.snapshots.accessible

jlsherrill:
I think currently there are two flags, enabled and accessible.  One details whether the feature is enabled in the environment, the second whether the user's org  has access.  you would think that if its not enabled it would not be accessible but this isn't actually the case.  I think we should change this on the backend and still use the accessible flag.  

So that if the feature is disable, it also would not be accessible.  That way you wouldn't have to check two different flags.  
Thoughts?

xbhouse:
@jsherrill that makes sense to me. where would it need to be changed in the backend?

jlsherrill:
I think you can just return false ere if feature.Enabled is false: 
https://github.com/content-services/content-sources-backend/blob/main/pkg/handler/features.go#L57-L68

xbhouse:
thanks @jlsherrill, getting expected behavior with features.snapshots.accessible now. should i change this PR back to accessible and create a PR for that change to the backend?

jlsherrill:
yep! sounds like a plan!

xbhouse:
@swadeley created this [PR](https://github.com/content-services/content-sources-backend/pull/457) for this change, so they'll need to be tested together

swadeley:
Hi

I deployed both PRs with snapshot feature disabled. The new text was not displayed (as expected). I than ran deployment command again with snapshot feature enabled.

In the image below, the first repo was created with snapshot=False, second with snapshot=True. The kebab menu also works as expected:

image

PR #189, opened 2023-12-11
Fixes 3274: Hides RH repo when snapshots not enabled

Andrewgdewar:
## Summary

- Makes it so that Red hat repositories will not be accessible without feature.snapshot permissions.

## Testing steps

- Turn off snapshotting in the backend config
- Run the ui, the toggle button for redhat repositories should no longer be accessible
- Deeplinking to content?origin=redhat will also not work.


jlsherrill:
https://issues.redhat.com/browse/HMS-3274

mshriver:
verified the components are hidden in the ephemeral, and its tied to FEATURES_SNAPSHOTS_ENABLED on the backend.

mshriver (approved):

PR #200, opened 2024-01-18
Fixes 1172: Clear Filter

kwarnerredhat:
## Testing steps-
"clicking clear filters button should remove any filters selected from popular repositories table"

### DIFF
--- a/src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
+++ b/src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
@@ -98,9 +98,7 @@ it('finds search box, enters text, and checks text occurrence', async () => {
   const searchBox = getByPlaceholderText('Filter by name/url');
   fireEvent.change(searchBox, { target: { value: 'yourSearchText' } });
 
-  waitFor(() => {
-    const occurrences = queryAllByText('yourSearchText', { exact: false });
-    expect(occurrences).toBeTruthy();
-    expect(occurrences.length).toBe(2);
-  });
+  const occurrences = queryAllByText('yourSearchText', { exact: false });
+  expect(occurrences).toBeTruthy();
+  expect(occurrences.length).toBe(2);
 });
--- a/src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
+++ b/src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
@@ -98,10 +98,5 @@ it('finds search box, enters text, and checks text occurrence', async () => {
   const searchBox = getByPlaceholderText('Filter by name/url');
   fireEvent.change(searchBox, { target: { value: 'yourSearchText' } });
 
-  await waitFor(() => {
-    return queryAllByText('yourSearchText', { exact: false }).length === 2;
-    const occurrences = queryAllByText('yourSearchText', { exact: false });
-    expect(occurrences).toBeTruthy();
-    expect(occurrences.length).toBe(2);
-  });
+  await waitFor(() => queryAllByText('yourSearchText', { exact: false }).length === 2);
 });
--- a/src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
+++ b/src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
@@ -98,7 +98,10 @@ it('finds search box, enters text, and checks text occurrence', async () => {
   const searchBox = getByPlaceholderText('Filter by name/url');
   fireEvent.change(searchBox, { target: { value: 'yourSearchText' } });
 
-  const occurrences = queryAllByText('yourSearchText', { exact: false });
-  expect(occurrences).toBeTruthy();
-  expect(occurrences.length).toBe(2);
+  await waitFor(() => {
+    return queryAllByText('yourSearchText', { exact: false }).length === 2;
+    const occurrences = queryAllByText('yourSearchText', { exact: false });
+    expect(occurrences).toBeTruthy();
+    expect(occurrences.length).toBe(2);
+  });
 });
--- a/src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
+++ b/src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
@@ -89,7 +89,7 @@ it('finds search box, enters text, and checks text occurrence', async () => {
     data: { data: [defaultPopularRepository], meta: { count: 1, limit: 20, offset: 0 } },
   }));
 
-  const { getByPlaceholderText, queryAllByText } = render(
+  const { getByPlaceholderText, queryAllByText, getByText, queryByText } = render(
     <ReactQueryTestWrapper>
       <PopularRepositoriesTable />
     </ReactQueryTestWrapper>,
@@ -98,5 +98,17 @@ it('finds search box, enters text, and checks text occurrence', async () => {
   const searchBox = getByPlaceholderText('Filter by name/url');
   fireEvent.change(searchBox, { target: { value: 'yourSearchText' } });
 
-  await waitFor(() => queryAllByText('yourSearchText', { exact: false }).length === 2);
+  await waitFor(() => {
+    const occurrences = queryAllByText('yourSearchText', { exact: false });
+    console.log('Rendered occurrences:', occurrences);
+    expect(occurrences.length).toBe(1);
+
+    const clearFiltersButton = getByText('Clear filters');
+    fireEvent.click(clearFiltersButton);
+  });
+
+  await waitFor(() => {
+    const chip = queryByText('yourSearchText');
+    expect(chip).toBeNull();
+  });
 });
--- a/src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
+++ b/src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
@@ -1,4 +1,4 @@
-import { fireEvent, render, waitFor } from '@testing-library/react';
+import { render, waitFor } from '@testing-library/react';
 import {
   ReactQueryTestWrapper,
   defaultPopularRepository,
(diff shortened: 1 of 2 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-1172

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.tsx
```diff
@@ -128,6 +137,12 @@ const PopularRepositoriesTable = () => {
   const debouncedSearchValue = useDebounce(searchValue);
   const [perPage, setPerPage] = useState(storedPerPage);
   const [isActionOpen, setIsActionOpen] = useState(false);
+  const [filterData, setFilterData] = useState<FilterData>({
+    searchQuery: '',
+    versions: [],
+    arches: [],
+    statuses: [],
+  });
```
Andrewgdewar, line 145: this is not necessary, we are only using searchValue.

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.tsx
```diff
@@ -343,6 +358,10 @@ const PopularRepositoriesTable = () => {
 
   const countIsZero = !data?.data?.length;
 
+  const clearFilters = () =>
+    setFilterData({ searchQuery: '', versions: [], arches: [], statuses: [] });
+  setSearchValue('');
```
Andrewgdewar, line 363: what does this line of code do.

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.tsx
```diff
@@ -343,6 +358,10 @@ const PopularRepositoriesTable = () => {
 
   const countIsZero = !data?.data?.length;
 
+  const clearFilters = () =>
+    setFilterData({ searchQuery: '', versions: [], arches: [], statuses: [] });
```
Andrewgdewar, line 362: Remove this, not needed.

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.tsx
```diff
@@ -466,6 +485,17 @@ const PopularRepositoriesTable = () => {
               isCompact
               onPerPageSelect={onPerPageSelect}
             />
+            <Hide
+              hide={
+                !debouncedSearchValue &&
+                filterData.versions.length === 0 &&
+                filterData.arches.length === 0
```
Andrewgdewar, line 492: Remove this, not needed.

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.tsx
```diff
@@ -597,7 +627,7 @@ const PopularRepositoriesTable = () => {
       </Hide>
       <Hide hide={!countIsZero}>
         <EmptyTableState
-          clearFilters={() => setSearchValue('')}
+          clearFilters={clearFilters}
```
Andrewgdewar, line 630: undo this change.

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.tsx
kwarnerredhat: I added that from the filters on the Content List Filters following the pattern but realized what you told me earlier

Andrewgdewar: This line of code is not a part of the "clearFilters" function above. It is separate. 
What does this line of code do by itself here in such a case?

kwarnerredhat: I removed the line it worked

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.tsx
```diff
@@ -452,6 +474,18 @@ const PopularRepositoriesTable = () => {
...
+              <Chip key='search_chip' onClick={() => setSearchQuery('')}>
+                {searchQuery}
+              </Chip>
+            </ChipGroup>
+          )}
+          {(debouncedSearchValue !== '' && searchQuery !== '') || (
+            <Button className={classes.clearFilters} onClick={clearFilters} variant='link' isInline>
+              Clear filters
+            </Button>
+          )}
```
Andrewgdewar, line 488: This code always shows, you need to hide the clear filters button when no search query is made.

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.tsx
```diff
@@ -452,6 +474,18 @@ const PopularRepositoriesTable = () => {
               </FlexItem>
             </InputGroupItem>
           </InputGroup>
+          {searchQuery !== '' && (
+            <ChipGroup categoryName='Name/URL'>
+              <Chip key='search_chip' onClick={() => setSearchQuery('')}>
+                {searchQuery}
+              </Chip>
+            </ChipGroup>
```
Andrewgdewar, line 482: This chip isn't showing yet.

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
```diff
@@ -30,54 +30,66 @@ it('expect PopularRepositoriesTable to render with add one item', () => {
     data: { data: [defaultPopularRepository], meta: { count: 1, limit: 20, offset: 0 } },
```
Andrewgdewar, line 30: In this file, you have changed a lot of code that doesn't need to be change.
Revert all changes, and either: 

A. **Add** your own test

B. **Add** code to an existing test. 

Do not delete/move any code that currently exists here please.

Andrewgdewar:
Below is a comparison between your current implementation and what is currently done on the "Customs repository" page:

Screenshot 2024-01-23 at 11 13 40 AM

Screenshot 2024-01-23 at 11 12 25 AM

Notice the padding/style differs quite a bit. 

Please adjust your implementation to match.

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
```diff
@@ -81,3 +81,25 @@ it('Render a loading state checking search disabled', () => {
...
+      <PopularRepositoriesTable />
+    </ReactQueryTestWrapper>,
+  );
+
+  const searchBox = getByPlaceholderText('Filter by name/url');
+  fireEvent.change(searchBox, { target: { value: 'yourSearchText' } });
+
+  waitFor(() => {
+    const occurrences = queryByText('yourSearchText', { exact: false });
+    expect(occurrences).toBeTruthy();
```
Andrewgdewar, line 103: This is starting to look good!  
But you need to find all "occurrences". In this case I believe you can use the "query**All**ByText". 

That method would return an array of items that are a match.

Currently if there is more than one item found you'd likely get an error that the test doesn't know which one you are wanting to target.

So once you have an array of items, you can check the length of that array (array.length) and have the expect look for a length of 2. 
There are a couple ways to do that with "expect" so I'll leave that for you to figure out.

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
```diff
@@ -81,3 +81,26 @@ it('Render a loading state checking search disabled', () => {
...
+    </ReactQueryTestWrapper>,
+  );
+
+  const searchBox = getByPlaceholderText('Filter by name/url');
+  fireEvent.change(searchBox, { target: { value: 'yourSearchText' } });
+
+  waitFor(() => {
+    const occurrences = queryAllByText('yourSearchText', { exact: false });
+    expect(occurrences).toBeTruthy();
+    expect(occurrences.length).toBe(2);
```
Andrewgdewar, line 104: Question, does this test still pass if you change this number to 1 or 4 or 88?

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
```diff
@@ -81,3 +81,26 @@ it('Render a loading state checking search disabled', () => {
...
+    <ReactQueryTestWrapper>
+      <PopularRepositoriesTable />
+    </ReactQueryTestWrapper>,
+  );
+
+  const searchBox = getByPlaceholderText('Filter by name/url');
+  fireEvent.change(searchBox, { target: { value: 'yourSearchText' } });
+
+  waitFor(() => {
+    const occurrences = queryAllByText('yourSearchText', { exact: false });
```
Andrewgdewar, line 102: if you change this "yourSearchText" to "somethingElse" will this test still pass?

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
```diff
@@ -81,3 +81,26 @@ it('Render a loading state checking search disabled', () => {
...
+    expect(occurrences).toBeTruthy();
```
Andrewgdewar, line 103: Read [this](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) on what "truthy means" (it just means not negative value) 
Query all by text returns an empty array "[]". 
Screenshot 2024-01-24 at 11 50 04 AM

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
kwarnerredhat: Yeah it still passs

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
kwarnerredhat: Yeah it still pass

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
kwarnerredhat: It prints out an array in the boolean context and highlights if its true

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
```diff
@@ -81,3 +81,26 @@ it('Render a loading state checking search disabled', () => {
...
+  const { getByPlaceholderText, queryAllByText } = render(
+    <ReactQueryTestWrapper>
+      <PopularRepositoriesTable />
+    </ReactQueryTestWrapper>,
+  );
+
+  const searchBox = getByPlaceholderText('Filter by name/url');
+  fireEvent.change(searchBox, { target: { value: 'yourSearchText' } });
+
+  waitFor(() => {
```
Andrewgdewar, line 101: waitFor is only used with async methods, and in this case needs to "await" ed.

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
Andrewgdewar: Reread the above

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
```diff
@@ -81,3 +81,22 @@ it('Render a loading state checking search disabled', () => {
...
+  const { getByPlaceholderText, queryAllByText } = render(
+    <ReactQueryTestWrapper>
+      <PopularRepositoriesTable />
+    </ReactQueryTestWrapper>,
+  );
+
+  const searchBox = getByPlaceholderText('Filter by name/url');
+  fireEvent.change(searchBox, { target: { value: 'yourSearchText' } });
+
+  await waitFor(() => queryAllByText('yourSearchText', { exact: false }).length === 2);
```
Andrewgdewar, line 101: this doesn't test anything you need to use the "expect" methods in tests.

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.test.tsx
```diff
@@ -81,3 +81,28 @@ it('Render a loading state checking search disabled', () => {
 
   expect(queryByPlaceholderText('Filter by name/url')).toHaveAttribute('disabled');
 });
+
+it('finds search box, enters text, and checks text occurrence', async () => {
```
Andrewgdewar, line 85: this test does not need to be async

kwarnerredhat: I removed it

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.tsx
```diff
@@ -125,6 +136,7 @@ const PopularRepositoriesTable = () => {
   const storedPerPage = Number(localStorage.getItem(perPageKey)) || 20;
   const [page, setPage] = useState(1);
   const [searchValue, setSearchValue] = useState('');
+  const [searchQuery, setSearchQuery] = useState('');
```
Andrewgdewar, line 139: We already have a value that stores the search text, it is called "searchValue" 
This extra state is not needed.

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.tsx
```diff
@@ -343,6 +355,11 @@ const PopularRepositoriesTable = () => {
 
   const countIsZero = !data?.data?.length;
 
+  const clearFilters = () => {
+    setSearchValue('');
+    setSearchQuery('');
```
Andrewgdewar, line 360: as mentioned above, this does the same thing.

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.tsx
```diff
@@ -343,6 +355,11 @@ const PopularRepositoriesTable = () => {
 
   const countIsZero = !data?.data?.length;
 
+  const clearFilters = () => {
```
Andrewgdewar, line 358: As this is just setting the searchValue to "". 
Do we need a function for this? or can we call it directly.

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.tsx
```diff
@@ -452,6 +468,25 @@ const PopularRepositoriesTable = () => {
               </FlexItem>
             </InputGroupItem>
           </InputGroup>
+          {(debouncedSearchValue !== '' || searchQuery !== '') && (
```
Andrewgdewar, line 471: Look into what debounced values are. 
Would it make sense to wait for the debouncedSearchValue to change this? 
Try typing something in the searchbox, and then hit clear filters. 
Does the chip immediately disappear?

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.tsx
```diff
@@ -355,12 +372,11 @@ const PopularRepositoriesTable = () => {
...
               <TextInput
                 isDisabled={isLoading}
-                type='text'
                 id='search'
-                ouiaId='popular_filter_search'
+                ouiaId='filter_search'
                 placeholder='Filter by name/url'
-                value={searchValue}
-                onChange={(_event, val) => setSearchValue(val)}
+                value={searchQuery}
```
Andrewgdewar, line 378: you've changed this value to a new one you've created, but the API is still using the old one. 

This makes it so that when you type into the search box, the api is not filtered. 

So at the moment, the searchbox does not function as intended.

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.tsx
kwarnerredhat: So the function might be needed

Andrewgdewar: It's not. if you remove searchQuery/setSearchQuery and use searchValue/setSearchValue. 
Then replace this function "clearFilters" with ()=> setSearchValue('') 
You will have the same functionality, and you will actually be able to "search"

src/Pages/PopularRepositoriesTable/PopularRepositoriesTable.tsx
```diff
@@ -366,8 +366,9 @@ const PopularRepositoriesTable = () => {
             <InputGroupItem isFill>
               <TextInput
                 isDisabled={isLoading}
+                type='text'
                 id='search'
-                ouiaId='filter_search'
+                ouiaId='popular_filter_search'
```
Andrewgdewar, line 371: I add this back as QE uses this to target the input for tests. 
This is likely the reason that your CI-CD Build was failing.

kwarnerredhat: Okay

Andrewgdewar (approved):
Changes look. good.

swadeley:
Hi

its looks good:

image


and it works.

Thank you.

PR #203, opened 2024-01-25
Fixes 2726: Custom repositories missing zero state

xbhouse:
## Summary

Adds a zero state for when a user has no custom repositories added yet. 

This is the current zero state:

Screenshot 2024-01-25 at 12 10 07 PM

## Testing steps

- When a user enters the app and has no custom repositories, they should see the zero state
- Clicking the `Add custom repositories` button from the zero state component should route them to the modal to add a repository
- Once they add a repository, they should not see the zero state once they leave and visit the app again. In that case they should instead be routed to the Repositories page 
- If a user enters the modal to add a repository from the zero state, and closes the modal before adding a repository, the empty repository table state should be shown. If they refresh the page, they should see the zero state again. 
- If a user has repositories and then deletes them all, upon refresh or navigating away and back to the app they should see the zero state

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,45 @@
 import '@redhat-cloud-services/frontend-components-utilities/styles/_all';
 import 'react18-json-view/src/style.css';
 import NotificationsPortal from '@redhat-cloud-services/frontend-components-notifications/NotificationPortal';
+import { Bullseye, Spinner } from '@patternfly/react-core';
+import { useEffect, /* useMemo,*/ useState } from 'react';
+// import { useLocation } from 'react-router-dom';
+// import { last } from 'lodash';
+
 import Routes from './Routes';
 import { useAppContext } from './middleware/AppContext';
 import { NoPermissionsPage } from './components/NoPermissionsPage/NoPermissionsPage';
-
-import { Bullseye, Spinner } from '@patternfly/react-core';
+import { ContentOrigin, FilterData } from './services/Content/ContentApi';
+import { useContentListQuery } from './services/Content/ContentQueries';
+import { perPageKey } from './Pages/ContentListTable/ContentListTable';
+// import { REPOSITORIES_ROUTE } from './Routes/constants';
 
 export default function App() {
-  const { rbac, isFetchingFeatures } = useAppContext();
+  const { rbac, isFetchingFeatures, zeroState, setZeroState } = useAppContext();
+  const storedPerPage = Number(localStorage.getItem(perPageKey)) || 20;
+  // const { pathname } = useLocation();
+  // const isDefaultRoute = useMemo(() => last(pathname.split('/')) === REPOSITORIES_ROUTE, [pathname]);
+
+  const [filterData] = useState<FilterData>({
+    searchQuery: '',
+    versions: [],
+    arches: [],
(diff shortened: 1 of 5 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-2726

swadeley:
Hi

if you click around a bit and go back then the text `Don't have a RHEL subscription? Start a trial.` moves to the right.

image

Andrewgdewar:
> Hi
> 
> if you click around a bit and go back then the text `Don't have a RHEL subscription? Start a trial.` moves to the right.
> 
> image

This is caused by an internal component issue. 
The css used to define the box surrounding the Add custom query button in hardcoded to 300px max-height. 
The "Get started with Insights..." customText we have added, when wrapped causes the height to exceed this value, forcing a second column to be used.
Simply increasing the max-height to 320px, or making the box height variable would solve the issue in our case.

Although we could create a css override, a better fix would be to have the owner of the component fix this globally, or make the internal facing classes public via a variable on the "AsyncComponent" itself. 

@xbhouse a temporary fix for this issue would be to add the following to your minHeight class:
`
  minHeight: {
    minHeight: '100%',
    '& .bannerBefore': {
      maxHeight: '320px!important',
    },
  },
`

swadeley:
Hi, the repo page looks good now

image


Thank you

swadeley:
Hi @xbhouse 

It says PF4 not PF5 for the component type:

<h1 data-ouia-component-type="PF4/Title" data-ouia-safe="true" data-ouia-component-id="ZeroStateCustomAppTitle" style="color: rgb(21, 21, 21);" class="pf-c-title pf-m-2xl">Start using Content management now</h1>

xbhouse:
@swadeley oh yep looks like they're using PF4 [here](https://github.com/RedHatInsights/insights-dashboard/blob/7f3c968826eaa8cbdef952eb201c318d518176f6/package.json#L64). @Andrewgdewar does this cause an issue if we're using PF5 elsewhere?

swadeley:
Hi

I see error in UI:
`y.useLastVisited() is null`

Andrewgdewar (approved):
Looks good! Ack!

src/App.tsx
```diff
@@ -1,17 +1,45 @@
...
+import { useContentListQuery } from './services/Content/ContentQueries';
+import { perPageKey } from './Pages/ContentListTable/ContentListTable';
+// import { REPOSITORIES_ROUTE } from './Routes/constants';
 
 export default function App() {
-  const { rbac, isFetchingFeatures } = useAppContext();
+  const { rbac, isFetchingFeatures, zeroState, setZeroState } = useAppContext();
+  const storedPerPage = Number(localStorage.getItem(perPageKey)) || 20;
+  // const { pathname } = useLocation();
+  // const isDefaultRoute = useMemo(() => last(pathname.split('/')) === REPOSITORIES_ROUTE, [pathname]);
```
xbhouse, line 21: @adewar once i rebased some of this logic broke (the zero state page would show up even if there were repositories). it works now that i've commented it out, but i know we added this to make sure the useContentListQuery wasn't being called when it didn't need to be. let me know if you have any ideas on how to add this functionality back in

PR #206, opened 2024-01-31
Fixes 3472: change snapshot delta to only use package count

jlsherrill:
## Summary

Errata delta in pulp is unreliable, as ordering of things seem to count as a new errata, throwing the counts off.  For now we will just show package change count.

## Testing steps

Snapshot a repo with package and errata (https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/) , view delta on repos table and snapshot list.  Delta count should match package count.

### DIFF
--- a/src/testingHelpers.tsx
+++ b/src/testingHelpers.tsx
@@ -202,8 +202,8 @@ export const defaultSnapshotItem: SnapshotItem = {
     'rpm.packagegroup': 20,
   },
   added_counts: {
-    'rpm.advisory': 3864,
-    'rpm.package': 17208,
+    'rpm.advisory': 100,
+    'rpm.package': 200,
     'rpm.packagecategory': 1,
     'rpm.packageenvironment': 1,
     'rpm.packagegroup': 20,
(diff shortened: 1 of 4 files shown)


src/testingHelpers.tsx
```diff
@@ -202,8 +202,8 @@ export const defaultSnapshotItem: SnapshotItem = {
     'rpm.packagegroup': 20,
   },
   added_counts: {
-    'rpm.advisory': 3864,
-    'rpm.package': 17208,
+    'rpm.advisory': 100,
+    'rpm.package': 200,
```
jlsherrill, line 206: the tests failed because these numbers showed up twice in the table (one for delta, one for package count).  I just changed the numbers to match, but an alternative would be to find the element more precisely.

jlsherrill:
https://issues.redhat.com/browse/HMS-3472

Andrewgdewar (approved):

Andrewgdewar:
@swadeley :  This may result in a change to test automation if your tests are expecting a specific number value from the "Changed" number for snapshots in the repoList view or snapshotList view

The current test failures appear to be unrelated, so I am unsure if the above will be a problem.

PR #213, opened 2024-02-06
Move bootstrap.sh after frontend-build.sh

swadeley:
## Summary

Prevent `bootstrap.sh`  from setting a commit based tag + context based tag by loading it later when  IMAGE_TAG is already set.

### DIFF
--- a/.rhcicd/pr_check.sh
+++ b/.rhcicd/pr_check.sh
@@ -33,12 +33,12 @@ export DEPLOY_FRONTENDS="true"
 
 set -exv
 
+# source is preferred to | bash -s in this case to avoid a subshell
+source <(curl -sSL $COMMON_BUILDER/src/frontend-build.sh)
+
 # bootstrap bonfire and it's config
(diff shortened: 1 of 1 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-3558

mshriver:
I've raised this with service delivery, because the cicd-tools bootstrap doesn't check if IMAGE_TAG is already set.

It has the same logic for image tagging that frontend-builder-common does, so it should land on the same tag, but this may need further work. This will at least stop it from pushing tags with duplications in them.

jlsherrill:

swadeley:
@mshriver Can we merge this?

mshriver (commented):
This does require an update to frontend-build or cicd-tools bootstrapping.  It works around things, but its possible the image naming will change and then the deploy will break again. 

https://issues.redhat.com/browse/RHCLOUD-30865

mshriver (approved):

swadeley:
Thank you

PR #217, opened 2024-02-19
Fixes 3630: hide snapshotting in prod stable

jlsherrill:
## Summary

Hides snapshotting in prod stable

## Testing steps

Turn on snapshotting in the backend:

```
features:
  snapshots:
    enabled: true
    accounts: 
    users: 
```

launch frontend with stage stable/beta, prod stable/beta.

It should not be available in prod stable, but should be elsewhere.

### DIFF
--- a/src/middleware/AppContext.tsx
+++ b/src/middleware/AppContext.tsx
@@ -39,6 +39,12 @@ export const ContextProvider = ({ children }: { children: ReactNode }) => {
 
     (async () => {
       const fetchedFeatures = await fetchFeatures();
+      // Disable snapshotting in prod stable
+      if (chrome.isProd() && !chrome.isBeta()) {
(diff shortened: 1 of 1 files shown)


src/middleware/AppContext.tsx
```diff
@@ -39,6 +39,12 @@ export const ContextProvider = ({ children }: { children: ReactNode }) => {
 
     (async () => {
       const fetchedFeatures = await fetchFeatures();
+      // Disable snapshotting in prod stable
+      if (chrome.isProd() && !chrome.isBeta()) {
```
Andrewgdewar, line 43:
```suggestion
      if (chrome.isProd() && !chrome.isBeta() && fetchedFeatures?. snapshots?.accessible) {
```

Andrewgdewar (approved):
One comment, the rest looks good.

PR #220, opened 2024-02-27
Fixes 3653: manual snapshot trigger disabled for specific repos

xbhouse:
## Summary

- Disables the `Trigger snapshot` action if a repo does not have snapshotting enabled and adds a tooltip to alert the user

## Testing steps

- Create a repository without snapshotting enabled
- Upon clicking the kebab for that repository, you should see the `Trigger snapshot` action disabled and a tooltip on that action that says `Snapshots disabled for this repository.`


src/Pages/ContentListTable/ContentListTable.tsx
```diff
@@ -282,8 +283,15 @@ const ContentListTable = () => {
...
                   {
-                    isDisabled: actionTakingPlace || rowData?.status === 'Retrying',
-                    title: 'Trigger snapshot',
+                    isDisabled:
+                      actionTakingPlace || rowData?.status === 'Retrying' || !rowData.snapshot,
+                    title: rowData.snapshot ? (
+                      'Trigger snapshot'
+                    ) : (
+                      <Tooltip content='Snapshots disabled for this repository.' isVisible>
+                        <p>Trigger snapshot</p>
```
xbhouse, line 292: not sure if this is the best way to add a tooltip here. thoughts?

jlsherrill:
https://issues.redhat.com/browse/HMS-3653

src/Pages/ContentListTable/ContentListTable.tsx
Andrewgdewar: I think theres a custom component for this?
ConditionalTooltip

Andrewgdewar: Although you'd have to add a <Text> component or similar to ensure the isDisabled prop is being passed to something that can use it.

src/Pages/ContentListTable/ContentListTable.tsx
```diff
@@ -282,11 +282,17 @@ const ContentListTable = () => {
...
+                      actionTakingPlace || rowData?.status === 'Retrying' || !rowData.snapshot,
                     title: 'Trigger snapshot',
                     onClick: () => {
                       triggerSnapshot(rowData.uuid);
                     },
+                    tooltipProps: {
+                      content: 'Snapshots disabled for this repository.',
+                      isVisible: !rowData.snapshot,
+                      trigger: 'manual',
+                    },
```
Andrewgdewar, line 295: So I tested this out, and although it will present the tooltip, the tooltip isn't being triggered via mouseenter / focus. 
So when I open the menu the tooltip would appear. 

I played around with this and found that you can use the "triggerRef" to attach the tooltip to the document programmatically. 

You can also use the !rowData.snapshot bool to conditionally add that value to the row. 

Here's what I ended up with.

`
tooltipProps: !rowData.snapshot
  ? {
      content: 'Snapshots disabled for this repository.',
      triggerRef: () =>
        document.getElementById('actions-column') || document.body,
    }
  : undefined,
`
`
   <ActionsColumn id='actions-column' items={rowActions(rowData)} />
`

xbhouse: oh yep you're right, i was moving my mouse too fast so i thought it was just always happening on mouseenter XD will try this, thank you!

Andrewgdewar (approved):
Ack! 
Optionally: Consider having the tooltip fall left so that it does not block other items from being clicked.

xbhouse:
ooh ok i haven't seen it pop up anywhere other than to the left yet, i'll add a position to make sure it's always left

mayurilahane:
going to merge it first and then test it 
Thanks!

PR #222, opened 2024-02-28
Fixes 2937,2956: SnapshotListModal with snapshot selector 

Andrewgdewar:
## Summary

This adds the snapshot detail modal with working snapshot selector, search, and back button. 

## Testing steps
This PR is dependent on [this backend API](https://github.com/content-services/content-sources-backend/pull/590).

## Still Needs
- [x]  Tests


jlsherrill:
https://issues.redhat.com/browse/HMS-2937

jlsherrill:
https://issues.redhat.com/browse/HMS-2956

xbhouse:
the link on the package count from the snapshot list modal seems to route me back to the custom repos view instead of the snapshot detail modal, does that happen for you?

src/Pages/ContentListTable/components/SnapshotDetailsModal/SnapshotDetailsModal.tsx
```diff
@@ -0,0 +1,97 @@
...
+    _: React.MouseEvent<HTMLElement, MouseEvent>,
+    tabIndex: string | number,
+  ) => {
+    setActiveTabKey(tabIndex);
+  };
+
+  const onClose = () =>
+    navigate(rootPath + (contentOrigin === ContentOrigin.REDHAT ? `?origin=${contentOrigin}` : ''));
+
+  const onBackClick = () => navigate(rootPath + `/${repoUUID}/snapshots`);
```
xbhouse, line 50: this should probably be `/${REPOSITORIES_ROUTE}/${repoUUID}/snapshots` too, i think that's the last one

src/Pages/ContentListTable/components/SnapshotDetailsModal/Tabs/SnapshotPackagesTab.tsx
```diff
@@ -0,0 +1,216 @@
...
+        <Grid className={classes.mainContainer}>
+          <SkeletonTable
+            rows={perPage}
+            numberOfColumns={columnHeaders.length}
+            variant={TableVariant.compact}
+          />
+        </Grid>
+      </Hide>
+      <Hide hide={fetchingOrLoading}>
+        <Table aria-label='Custom repositories table' ouiaId='packages_table' variant='compact'>
```
xbhouse, line 167: i think this aria-label should be `snapshot_packages_table` or something similar

xbhouse:
it seems like the columns in the snapshot detail modal are not sortable (though the arrows seem to indicate they should be). actually.. it looks like the column sort isn't working for the custom repos table now too so not specific to this PR, any idea why that might be?

src/Pages/ContentListTable/components/SnapshotDetailsModal/SnapshotDetailsModal.tsx
Andrewgdewar: Thanks for catching all of these!

Andrewgdewar:
> it seems like the columns in the snapshot detail modal are not sortable (though the arrows seem to indicate they should be). actually.. it looks like the column sort isn't working for the custom repos table now too so not specific to this PR, any idea why that might be?

I did recently fix the custom repos issue somewhere for the sort. I'll check if it's the same issue here (no deps in the sortBy useMemo).

Andrewgdewar:
So checking the not sorting business. 
I merged a fix for the content-list sort yesterday I think (on main) so will pull that in. 
The package list not sorting. Perhaps @jlsherrill's API doesn't support that like the old package list did. 
Looking into that now.

swadeley:
Hi, I see "json: unknown field \"pulp_last_updated\""}]} in the pr check.
 I think we saw that on another PR recently. I wonder if a rebase would clear that error.

PR #230, opened 2024-03-08
MergeConflictSortFix

Andrewgdewar:
## Summary

Introduced a bug that prevented sorting on the repositories list view table 3 days ago. 

This fixes it.


swadeley (approved):
ack

PR #234, opened 2024-03-20
Fixes 3473: snapshot status update

xbhouse:
## Summary

- Updates ContentListTable's StatusIcon to use the unified status for introspection and snapshotting

## Testing steps

- Needs to be tested with this [backend PR](https://github.com/content-services/content-sources-backend/pull/587)
- Introspect and snapshot a few repositories 
- The repository's unified status should be reflected in the UI instead of the last introspection status
- If a repository's status is In progress and the snapshot task is running, you should see a tooltip on the status with content `Repository snapshot in progress`. Otherwise if the status is pending you should see `Repository is being introspected`
- If a repository's status is Invalid and there is a snapshot error, you should see that error in the tooltip. Otherwise, if there's an introspection error, you should see that one. If neither has an error, you should see `An unknown error occurred`
- Same as above if a repository's status is Unavailable

### DIFF
--- a/src/Pages/ContentListTable/ContentListTable.tsx
+++ b/src/Pages/ContentListTable/ContentListTable.tsx
@@ -541,10 +541,10 @@ const ContentListTable = () => {
                         distribution_arch,
                         distribution_versions,
                         last_introspection_time,
-                        status
+                        status,
                       } = rowData;
                       return (
-                        <Tr key={uuid+status}>
+                        <Tr key={uuid + status}>
                           <Hide hide={!rbac?.write || isRedHatRepository}>
                             <Td
                               select={{
--- a/src/Pages/ContentListTable/ContentListTable.tsx
+++ b/src/Pages/ContentListTable/ContentListTable.tsx
@@ -407,6 +407,19 @@ const ContentListTable = () => {
     throw new Error('Unable to load Red Hat repositories');
   }
 
+  const showPendingTooltip = (
+    snapshotStatus: string | undefined,
+    introspectStatus: string | undefined,
+  ) => {
+    if (!snapshotStatus && !introspectStatus) {
+      return 'Introspection or snapshotting is in progress';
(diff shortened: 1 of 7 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-3473

Andrewgdewar:
Tested, I noticed the following when I go to snapshot a repository.
Screenshot 2024-03-25 at 12 23 19 PM
Screenshot 2024-03-25 at 12 23 16 PM

I did confirm that the snapshot was running.

xbhouse:
good catch @Andrewgdewar! recent [backend](https://github.com/content-services/content-sources-backend/pull/587) push should fix this

xbhouse:
@Andrewgdewar i'm seeing a couple more issues here:
- the status is sometimes blank on creation of a repo or triggering a snapshot or introspection (i think it's defaulting to the last case in the StatusIcon where it returns <></>). i can resolve this by making the default case 'Pending', but do you have any ideas why this may be happening or how to resolve this? 
- introspection data is available even if the repo is newly added and never been introspection (how can i clear this out?)

also, Stephen noticed that the tooltip doesn't change from saying introspecting to snapshotting when a repo is created with snaps enabled and that navigation is needed to move the status from in progress.

swadeley:
> also, @swadeley (correct me if i'm wrong) noticed these 2 things:
> 
>     * the tooltip doesn't change from saying introspecting to snapshotting when a repo is created with snaps enabled and it has completed introspection
> 
>     * navigation is needed to move the status from in progress (i see this intermittently if i add a console.log on the status weirdly, otherwise it's always pending until you refresh / navigate away and back / click on one of the columns)

Hi, the tooltip was correct for the Status but not for the kebab menu

Screenshot at 2024-03-28 11-28-18_1
Screenshot at 2024-03-28 11-28-25_2

Andrewgdewar (approved):

swadeley:
lgtm, thank you

PR #235, opened 2024-03-22
Fixes 2938: Add Errata tab to snapshot details

Andrewgdewar:
## Summary

- Adds a query to retrieve snapshot errata
- Adds an Errata tab to SnapshotDetails modal to list snapshot errata
- Supports filtering by name / synopsis, type, and severity
- Dependent on these PRs: [backend](https://github.com/content-services/content-sources-backend/pull/608), [tang](https://github.com/content-services/tang/pull/6)

## Testing steps

- Add a repository with package errata and let it snapshot 
- Click View all snapshots from the kebab on the Repositories page
- Click the Errata count to route to the Errata (Advisories) tab in the SnapshotDetails modal
- You should see all the errata for this snapshot displayed, with additional info for each errata displayed when clicking the expandable icon 
- Filtering on name / synopsis, type, severity should only display errata with those filters

### DIFF
--- a/src/components/SharedTables/AdvisoriesTable/index.tsx
+++ b/src/components/SharedTables/AdvisoriesTable/index.tsx
@@ -45,6 +45,7 @@ const useStyles = createUseStyles({
   rightMargin: { marginRight: '6px' },
   red: { extend: 'rightMargin', color: red },
   green: { extend: 'rightMargin', color: green },
+  retainSpaces: { whiteSpace: 'pre-line' },
 });
 
 interface Props {
@@ -159,7 +160,7 @@ export default function AdvisoriesTable({
                             </Flex>
                             <Grid>
                               <strong>Description</strong>
-                              <Text>{description}</Text>
+                              <Text className={classes.retainSpaces}>{description}</Text>
                             </Grid>
                             <Grid>
                               <strong>Reboot</strong>
--- a/src/components/SharedTables/AdvisoriesTable/index.tsx
+++ b/src/components/SharedTables/AdvisoriesTable/index.tsx
@@ -64,10 +64,11 @@ export default function AdvisoriesTable({
 }: Props) {
   const classes = useStyles();
   const columnHeaders = [
-    { name: 'Name' },
+    { name: 'Name', width: 15 },
     { name: 'Synopsis' },
     { name: 'Type', width: 15 },
     { name: 'Severity', width: 10 },
+    { name: 'Publish date', width: 15},
   ];
   const [expandState, setExpandState] = useState({});
 
@@ -140,6 +141,7 @@ export default function AdvisoriesTable({
                     <Td>
                       <SeverityWithIcon severity={severity} />
                     </Td>
+                    <Td>{formatDateDDMMMYYYY(issued_date)}</Td>
                   </Tr>
                   <Hide hide={!expandState[rowIndex]}>
                     <Tr>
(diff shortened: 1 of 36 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-2938

swadeley:
Hi

I added repo:
https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/

I clicked kebab, View all snapshots, clicked on the errata count "4", 
"Snapshot detail" table was displayed but the lines were blank.

Table closed and the follow error was displayed:

```
Unable to find errata with the given UUID.
Bad Request
```

xbhouse:
sorry just seeing this @swadeley. i wonder if the errors you're seeing are because the backend requires a new tang release. the [tang PR ](https://github.com/content-services/tang/pull/6) will likely need to be merged in first before this can be tested properly, otherwise you'd need to reference a local clone of that tang PR in the backend

Andrewgdewar:
✅  Ack on the code changes from me! 
Feel free to ack this if you are good with the other changes @xbhouse!
Thanks for the help 🙇‍♂️

src/components/SharedTables/AdvisoriesTable/index.tsx
rverdile, line 1: The JIRA card shows a "Published Date" column on the table, which might be the Issued Date?

image

Does that column need to be included here as well?

rverdile, line 1: The JIRA card says to include a link to the advisory page in the description, but I don't see it here. Should that be added or has the requirement changed?

image

xbhouse: ooh good point, it also looks like Reboot required is in the table as well instead of the description. @Andrewgdewar do you know if this mockup in the [JIRA](https://issues.redhat.com/browse/HMS-2938) is current? do we need to include these values in the table?

xbhouse: cc @Andrewgdewar, similar question as above

Andrewgdewar: I made the table identical to what is in patch today, the designs are kind of incomplete. 
@jlsherrill  can chime in here as I believe the idea was to keep it as close to that look as possible.

jlsherrill: we might need some funny logic that checks if the errata starts with RH and add a link if so

jlsherrill: probably to match patch:


image

swadeley:
Hi

looks good:

image


image

xbhouse:
ack on @Andrewgdewar's code changes!

rverdile:
I agree with justin's comment, otherwise looks good to me

jlsherrill:
Checking an epel errata, theres some formatting issues, for example:
image

jlsherrill:
otherwise everything else looks great!

xbhouse:
oh that doesn't look good at all :) there's no formatting being done on the description right now, i'll see what i can do there. which repo was that @jlsherrill ?

xbhouse:
for the sorting on date, i think we'd have to make a change in the api to support that. i can open a bug? correct me if i'm wrong @Andrewgdewar

jlsherrill:
We probably just need to adjust the default order here:  https://github.com/content-services/tang/blob/main/pkg/tangy/rpm.go#L313

as right now its not configurable, which i think is okay for now

jlsherrill:
and yeah, that can be a followup ticket

xbhouse:
sort order changed thanks to @Andrewgdewar ! and here's that epel errata description with the latest push @jlsherrill. thoughts? 

Screenshot 2024-04-25 at 3 27 53 PM

jlsherrill (approved):

PR #246, opened 2024-04-16
Fixes 3953: modularity filtering option is not clear

xbhouse:
## Summary

Fixes the modularity filtering toggle to default to ON if modularity filtering is enabled (default behavior) in repo create / edit modals

## Testing steps

- Add a repo with default modularity filtering behavior (toggle should be ON, label text should display `Modularity filtering enabled`, and should set `module_hotfixes = false` on the repository)
- Edit the repo to disable modularity filtering by toggling to OFF (label text should display `Modularity filtering disabled` and set `module_hotfixes = true`)

### DIFF
--- a/src/Pages/ContentListTable/components/AddContent/AddContent.tsx
+++ b/src/Pages/ContentListTable/components/AddContent/AddContent.tsx
@@ -488,7 +488,7 @@ const AddContent = () => {
               gpgLoading,
               metadataVerification,
               snapshot,
-              moduleHotfixesEnabled,
+              modularityFilteringEnabled,
             },
             index,
           ) => (
@@ -664,14 +664,14 @@ const AddContent = () => {
                       <Switch
                         label='Modularity filtering enabled'
                         labelOff='Modularity filtering disabled'
-                        ouiaId={`module_hotfixes_switch_${!moduleHotfixesEnabled ? 'on' : 'off'}`}
+                        ouiaId={`module_hotfixes_switch_${modularityFilteringEnabled ? 'on' : 'off'}`}
                         aria-label='enable_module_hotfixes'
                         hasCheckIcon
                         id={'module-hotfixes-switch' + index}
                         name='module-hotfixes-switch'
-                        isChecked={!moduleHotfixesEnabled}
+                        isChecked={modularityFilteringEnabled}
                         onChange={() => {
-                          updateVariable(index, { moduleHotfixesEnabled: !moduleHotfixesEnabled });
+                          updateVariable(index, { modularityFilteringEnabled: !modularityFilteringEnabled });
                         }}
                       />
                       <Tooltip content='Optional: Selecting this will set the module_hotfixes flag on the clients, allowing the repository to not be filtered by modularity'>
--- a/src/Pages/ContentListTable/components/AddContent/AddContent.tsx
+++ b/src/Pages/ContentListTable/components/AddContent/AddContent.tsx
@@ -674,7 +674,7 @@ const AddContent = () => {
                           updateVariable(index, { modularityFilteringEnabled: !modularityFilteringEnabled });
                         }}
                       />
-                      <Tooltip content='When enabled, modularity filtering prevents updates for packages contained within a module in this repository'>
+                      <Tooltip content='When enabled, modularity filtering prevents updates to packages contained within an enabled module'>
                         <OutlinedQuestionCircleIcon
                           className='pf-u-ml-xs'
                           color={global_Color_200.value}
--- a/src/Pages/ContentListTable/components/AddContent/AddContent.tsx
+++ b/src/Pages/ContentListTable/components/AddContent/AddContent.tsx
@@ -674,7 +674,7 @@ const AddContent = () => {
                           updateVariable(index, { modularityFilteringEnabled: !modularityFilteringEnabled });
                         }}
                       />
-                      <Tooltip content='Optional: Selecting this will set the module_hotfixes flag on the clients, allowing the repository to not be filtered by modularity'>
+                      <Tooltip content='When enabled, modularity filtering prevents updates for packages contained within a module in this repository'>
                         <OutlinedQuestionCircleIcon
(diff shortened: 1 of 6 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-3953

jlsherrill:
everything else looks great!

src/Pages/ContentListTable/components/AddContent/AddContent.tsx
```diff
@@ -662,14 +662,14 @@ const AddContent = () => {
...
+                        labelOff='Modularity filtering disabled'
+                        ouiaId={`module_hotfixes_switch_${!moduleHotfixesEnabled ? 'on' : 'off'}`}
                         aria-label='enable_module_hotfixes'
                         hasCheckIcon
                         id={'module-hotfixes-switch' + index}
                         name='module-hotfixes-switch'
-                        isChecked={moduleHotfixesEnabled}
+                        isChecked={!moduleHotfixesEnabled}
                         onChange={() => {
                           updateVariable(index, { moduleHotfixesEnabled: !moduleHotfixesEnabled });
```
Andrewgdewar, line 674: This is a lot of double negatives, for my 🧠 's sanity would you mind switching the name and reversing the defaults?  😆

swadeley:
> 
> currently the help text says:
> 
> ```
> Optional: Selecting this will set the module_hotfixes flag on the clients, allowing the repository to not be filtered by modularity
> ```
> 
> I might would say:
> 
> ```
> Disable filtering to enable package updates for packages contained within a module.  Sets the module_hotfixes="true" on clients if filtering is disabled.
> ```
> 
> @swadeley might have further suggestions :)

Hi
Re:
Optional: Selecting this will set the module_hotfixes flag on the clients, allowing the repository to not be filtered by modularity

 "Optional" is used in procedures in docs but is not required in a UI where other visual clues inform the user if something is required or not. In any case, tooltip space is valuable. Also, you are not supposed to describe GUIs in docs and the use of "Selecting this will set" seems like verbose description to me. 

"Select to set module_hotfixes on clients to enable unfiltered updates from a modular repository"

Re: Disable filtering to enable package updates for packages contained within a module.  Sets the module_hotfixes="true" on clients if filtering is disabled.

This "on clients if filtering is disabled" seems conflicting.

"Disables filtering of updates for packages contained within a module.  Sets the module_hotfixes="true" on clients "

To avoid describing the GUI and to rather tell the user what to do:
"Set to disable filtering of updates for packages contained within a module.  Sets the module_hotfixes="true" on clients "

src/Pages/ContentListTable/components/AddContent/AddContent.tsx
xbhouse: updated. let me know if that's any better

xbhouse:
@swadeley good to know, looks like some other tooltips in the modal have "Optional" as well, so we may have included that to be consistent. i can change this one to: "Set to disable filtering of updates for packages contained within a module. Sets module_hotfixes="true" on clients". @jlsherrill does that sound good?

src/Pages/ContentListTable/components/AddContent/AddContent.tsx
```diff
@@ -662,16 +662,16 @@ const AddContent = () => {
...
                         id={'module-hotfixes-switch' + index}
                         name='module-hotfixes-switch'
-                        isChecked={moduleHotfixesEnabled}
+                        isChecked={modularityFilteringEnabled}
                         onChange={() => {
-                          updateVariable(index, { moduleHotfixesEnabled: !moduleHotfixesEnabled });
+                          updateVariable(index, { modularityFilteringEnabled: !modularityFilteringEnabled });
                         }}
                       />
                       <Tooltip content='Optional: Selecting this will set the module_hotfixes flag on the clients, allowing the repository to not be filtered by modularity'>
```
ragevou, line 677: Hi @xbhouse and @jlsherrill . I suggest the following:
```suggestion
                      <Tooltip content='When enabled, the modularity filtering prevents package updates on this repository'>
```

swadeley:
> @swadeley good to know, looks like some other tooltips in the modal have "Optional" as well, so we may have included that to be consistent.

Consistency is a wonderful thing.

 i can change this one to: "Set to disable filtering of updates for packages contained within a module. Sets module_hotfixes="true" on clients". @jlsherrill does that sound good?

src/Pages/ContentListTable/components/AddContent/AddContent.tsx
swadeley: But its only filtering updates for the packages contained in the module. So, your suggestion I think should be:

'When enabled, modularity filtering prevents updates for packages contained within a module in this repository'

No need article because we mention "modularity filtering" as concept here, not a specific instance of it.

xbhouse: changed tooltip text to this suggestion: 'When enabled, modularity filtering prevents updates for packages contained within a module in this repository'

swadeley:
lgtm,over to @jlsherrill for ACK

jlsherrill:
a couple of small confusions:
```
When enabled, modularity filtering prevents updates for packages contained within a module in this repository
```

kinda implies two things:
1) the packages in this repository will be updated (by something?)  but in reality these packages are likely the ones providing the updates
2) this repository has modules, when in reality it probably doesn't.  Its probably the Appstream repo that does.  But because the appstream repo says 'foo' is in a module, and a system is using the 'foo' module, these are ignored, since they aren't in a module.

maybe something like: 
```
When enabled, modularity filtering prevents visibility of packages in this repository, when an enabled module contains these packages
```

Andrewgdewar (approved):
Code looks much clearer now, thank you!

jlsherrill:
yes, i like:
```
When enabled, modularity filtering prevents updates to packages contained within an enabled module.
```

xbhouse:
thanks all, changed to `When enabled, modularity filtering prevents updates to packages contained within an enabled module`

PR #263, opened 2024-05-10
Fixes 4105: Polling, css, console warn bugfix

Andrewgdewar:
## Summary
- Fixes a bug found with polling when navigating: When navigating to other pages/tabs and returning, polling would stop.  Now polling will continue regardless of page refresh, tab change, etc.. in cases where polling is needed (if there are pending states being returned).
- Patternfly has a known bug with their bottom-navigation Pagination component, I have applied css to temporarily fix this until it's patched
- Fixed a couple recurring console warns around the Th requiring an aria-label

## Testing steps
- Try before, try after.

### DIFF
--- a/src/AppEntry.tsx
+++ b/src/AppEntry.tsx
@@ -4,6 +4,7 @@ import { Provider as ReduxProvider } from 'react-redux';
 import * as Redux from 'redux';
 
 import App from './App';
+import 'index.scss';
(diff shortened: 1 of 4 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-4105

swadeley (approved):
lgtm

PR #279, opened 2024-06-11
Fixes 4257: expose/disable snaps by default in prod stable

xbhouse:
## Summary

- Exposes snapshotting in prod stable
- Temporarily disables snapshotting by default for custom and popular repos in prod stable

## Testing steps

When running the frontend on prod->stable:
- Snapshotting should be turned on
- When adding a custom repo, snapshot creation should default to disabled (see screenshot below)
Screenshot 2024-06-21 at 1 56 05 PM

- The top-level add / batch add button for popular repositories should add repositories with snapshotting disabled
- The hidden button in the add / batch add dropdown should show an option to "Add with snapshotting", which should add these repositories with snapshotting enabled (see screenshots below)
Screenshot 2024-06-21 at 1 55 44 PM
Screenshot 2024-06-21 at 1 55 34 PM

### DIFF
--- a/src/Pages/Repositories/PopularRepositoriesTable/PopularRepositoriesTable.tsx
+++ b/src/Pages/Repositories/PopularRepositoriesTable/PopularRepositoriesTable.tsx
@@ -62,6 +62,7 @@ import {
   DropdownToggle,
   DropdownToggleAction,
 } from '@patternfly/react-core/deprecated';
+import { useChrome } from '@redhat-cloud-services/frontend-components/useChrome';
 import { DELETE_ROUTE, REPOSITORIES_ROUTE } from 'Routes/constants';
 import { useHref, useNavigate } from 'react-router-dom';
 
@@ -145,6 +146,8 @@ const PopularRepositoriesTable = () => {
   const debouncedSearchValue = useDebounce(searchValue);
   const [perPage, setPerPage] = useState(storedPerPage);
   const [isActionOpen, setIsActionOpen] = useState(false);
+  const { isProd } = useChrome();
+  const isInProd = useMemo(() => isProd() === true, []);
 
   const onDropdownToggle = (_, isActionOpen: boolean) => {
     setIsActionOpen(isActionOpen);
@@ -342,6 +345,31 @@ const PopularRepositoriesTable = () => {
 
   const countIsZero = !data?.data?.length;
 
+  const dropdownItems = useMemo(() => {
+    if (isInProd) {
+      return [
+        <DropdownItem
+          key='action'
+          component='button'
+          onClick={() => addSelected(true)}
+          ouiaId='add-selected-repos-with-snap'
+        >
+          {`Add ${checkedRepositoriesToAdd.size} repositories with snapshotting`}
+        </DropdownItem>,
+      ];
+    } else
+      return [
+        <DropdownItem
+          key='action'
+          component='button'
+          onClick={() => addSelected(false)}
+          ouiaId='add-selected-repos-no-snap'
+        >
+          {`Add ${checkedRepositoriesToAdd.size} repositories without snapshotting`}
+        </DropdownItem>,
+      ];
+  }, [isInProd, checkedRepositoriesToAdd.size]);
+
   return (
     <Grid
       data-ouia-safe={!actionTakingPlace}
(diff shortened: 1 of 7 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-4257

src/Pages/Repositories/PopularRepositoriesTable/components/AddRepo.tsx
```diff
@@ -42,20 +44,64 @@ export const AddRepo = ({ isDisabled, addRepo }: Props) => {
...
-      data-ouia-component-id='add-popular_repo_without-snapshotting'
-      key='action'
-      component='button'
-      onClick={() => addRepo(false)}
-    >
-      Add without snapshotting
-    </DropdownItem>,
-  ];
+  const dropdownItems = (isProd: boolean) => {
+    if (isProd) {
```
Andrewgdewar, line 48: This all looks pretty good, what I'd suggest is checking isProd once with a useMemo above, as that value shouldn't change. 

Then using a constant for that value within the render, or a useMemo on this function with the isProd constant as a dependency.

The reason you'd make this change is to prevent unneeded re-renders of this component. 
Granted in this case, the total performance loss is likely very very small; the concept is more what I'm after for you to practice here, as performance difference on larger pages, or repeated components can be significant.

You can test the result of this described change by adding a console.log to your dropdownItems function here, or adding a count variable to the context, and allowing each of the dropdowns to trigger a single update of that count. This can make the results of the change much easier to see.

With the described method you should see far less re-renders.

src/Pages/Repositories/PopularRepositoriesTable/PopularRepositoriesTable.tsx
```diff
@@ -356,6 +358,30 @@ const PopularRepositoriesTable = () => {
 
   const countIsZero = !data?.data?.length;
 
+  const dropdownItems = (isProd: boolean, text: string) => {
+    if (isProd) {
```
Andrewgdewar, line 362: Similar as seen below.

src/Pages/Repositories/PopularRepositoriesTable/components/AddRepo.tsx
xbhouse: thanks! using a useMemo, it looks like the amount of renders were reduced by about 5 times. let me know if my changes are what you meant

src/Pages/Repositories/PopularRepositoriesTable/PopularRepositoriesTable.tsx
```diff
@@ -399,43 +430,69 @@ const PopularRepositoriesTable = () => {
...
+                                isDisabled={isDisabled}
+                              />
+                            }
+                            isOpen={isActionOpen}
                           >
-                            {withoutSnapshotText}
-                          </DropdownItem>
-                        </Dropdown>
-                      );
+                            {dropdownItems(defaultText)}
```
Andrewgdewar, line 462: Curious. Why have a text variable at all if above you are conditionally rendering your dropdown items?
You could easily simplify this a bit, also, instead of using the "isProd()" directly, you have a "isInProd" variable here that is stable. Although this likely should cause any issues, the recheck of the "isProd()" function could cause a few unneeded renders.

src/Pages/Repositories/PopularRepositoriesTable/components/AddRepo.tsx
```diff
@@ -42,48 +45,94 @@ export const AddRepo = ({ isDisabled, addRepo }: Props) => {
...
-            isDisabled={isDisabled}
-          />
-        }
-        isOpen={isActionOpen}
-      >
-        {dropdownItems}
-      </Dropdown>
-    );
+    // temporarily disable snapshotting by default for popular repos
+    if (isProd()) {
```
Andrewgdewar, line 76: Similar to my previous comment.

Andrewgdewar (approved):
A few more #nit comments, but otherwise this looks good.

src/Pages/Repositories/PopularRepositoriesTable/PopularRepositoriesTable.tsx
xbhouse: understood about using the isInProd directly, that makes sense! i'll make that change

regarding the text variable, if i just render the text `Add ${checkedRepositoriesToAdd.size} repositories with / without snapshotting` instead of passing in the text to the dropdownItems, the `checkedRepositoriesToAdd.size` variable always seems to be set to 0. i agree, i don't think i should need to pass in a variable here

xbhouse: oh i think i just need to add that value to the dep array :)

swadeley:
@xbhouse please rebase

mayurilahane:
Hi, I verified these details after merging this MR 

So, yes I found the the cases mentioned in the screenshots above, in the prod preview and prod env too.

just a single question @xbhouse 

Looking at the testing steps when you say "Snapshotting should be turned on" in first step and then you say "Snapshot creation should default to disabled" in the second step.

what does it mean? where should I look for it? Or is it something that can be found in UI ?

xbhouse:
> Hi, I verified these details after merging this MR
> 
> So, yes I found the the cases mentioned in the screenshots above, in the prod preview and prod env too.
> 
> just a single question @xbhouse
> 
> Looking at the testing steps when you say "Snapshotting should be turned on" in first step and then you say "Snapshot creation should default to disabled" in the second step.
> 
> what does it mean? where should I look for it? Or is it something that can be found in UI ?

hi @mayurilahane ! 

"Snapshotting should be turned on" just means that users should be able to use the snapshotting feature, so if you've verified that anything around snapshotting is accessible in prod then that should be good. 

"Snapshotting should default to disabled" means that when users go to add a custom repo, that toggle to enable / disable snapshot creation should default to off (shown in the first screenshot).

PR #288, opened 2024-06-21
Fixes 4327: expanding one list opens all lists

xbhouse:
## Summary

Fixes bug in the delete confirmation modal where if bulk deleting multiple repos that are associated with more than 3 templates, expanding one repo's list of associated templates would expand all lists

## Testing steps

- Add 2 repos, let them snapshot, and create at least 3 templates with each repo
- Try to bulk delete the repos in the UI and click `and X more` in the delete confirmation modal. It should only expand the list of templates selected


jlsherrill:
https://issues.redhat.com/browse/HMS-4327

jlsherrill (approved):

PR #296, opened 2024-07-05
Fixes 3898: add glitchtip error component

rverdile:
## Summary
Adds an error boundary component that, sends an error message to the backend API. The backend API will log the error so we get a glitchtip notification for it.

## Testing steps
1. Checkout https://github.com/content-services/content-sources-backend/pull/730
2. Replace the AddContent modal with something like `return <button onClick={methodDoesNotExist}> Ahhhh </button>`.
3. The frontend UI should show a dropdown with the error (same behavior as main)
4. The backend should have a log for the error message you see in the UI


rverdile:
closing in favor of waiting for official support

PR #305, opened 2024-07-24
Fixes 4474: fix snapshot behavior in prod-beta

xbhouse:
## Summary

In prod-beta, snapshotting should not be disabled by default. This fixes that behavior 

## Testing steps
- In prod-beta, snapshotting should be turned on and enabled by default when adding a custom repository
- The button behavior for adding popular repositories should be the same as it is in stage (split button should be Add > Add without snapshotting instead of Add > Add with snapshotting)


Andrewgdewar (approved):
Ack!

jlsherrill:
https://issues.redhat.com/browse/HMS-4474

PR #307, opened 2024-07-26
Fixes 4223: Ensure modals do not cause redirect on close

Andrewgdewar:
## Summary

- Fixes an issue where modals were redirecting on close, causing a rerender of the root node > clearing filters and component state.

## Testing steps

- Set a filter on the repositories page, open and close any modal, the filter should remain.


xbhouse (approved):
looks good! ack!

PR #318, opened 2024-08-12
Fixes 3938: Fix admin task polling

Andrewgdewar:
##Summary
Admin task polling wasn't working consistently, this change simply applies the same fix that was done for the repositories table on the AdminTask table.

##Testing steps
- Take some action that will produce a task (snapshot, etc..)
- Check the admin task table, if any of the tables statuses appears as "running" or "pending", the api will be polled every 15 seconds; until all are settled.


jlsherrill:
https://issues.redhat.com/browse/HMS-3938

xbhouse (approved):
works well and looks good! ack!

mayurilahane:
created a repo with a snapshot and then edited the same repo to trigger the admin task logs.

It gave me `pending` status at first but just within a seconds all statutes were in `completed` status.

image


Screenshot from 2024-08-14 18-25-53

PR #344, opened 2024-09-19
Fixes 4670: show last error and status for templates

xbhouse:
## Summary

- Adds a status column to the template list page
- Displays any errors if there's an Invalid status

### Additional changes
- Also adds 2 commands to run the app against stage or prod (`npm run start:stage` and `npm run start:prod`) without needing to select from a menu - thank you @marusak for the suggestion!

## Testing steps

1. Add a custom and RH repository, let them snapshot, and create a template using latest content. The template status in the UI should change to In progress and then Valid once the update-template-content task completes.
2. Edit the custom repo URL used in the template to create a new snapshot. Once the snapshot finishes, stop the server and remove the pulp URL. Restart the server. The template status should now be Invalid and display the error from the update-latest-snapshot task when clicking on it.
3. Edit the template. The status should still be Invalid, with the error from the update-template-content task displayed on click.

For QE - it would be difficult to simulate an error here, so validating the happy path should be sufficient. I've added a check to see if we're in the ephemeral environment, so in ephemeral it will not depend on the task that only runs with candlepin enabled

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,11 @@ Alternativly, simply add these lines to your /etc/hosts:
 
 4. `npm run build` (only required when setting up for the first time)
 
-5. `npm run start` to run against stage or prod environments. <br/>
+5. `npm run start` to choose whether to run against stage or prod environments. <br/>
+   OR <br/>
+   `npm run start:stage` to run against stage environment. <br/>
+   OR <br/>
+   `npm run start:prod` to run against prod environment. <br/>
    OR <br/>
    `npm run local` to run against a local backend running on port 8000. <br/>
    (keep in mind that you have to be connected to the VPN for this to work, even in the offices)
--- a/src/Pages/Templates/TemplatesTable/TemplatesTable.tsx
+++ b/src/Pages/Templates/TemplatesTable/TemplatesTable.tsx
@@ -297,9 +297,21 @@ const TemplatesTable = () => {
                 </Tr>
               </Thead>
               <Tbody>
-                {templateList.map((rowData: TemplateItem, index) => {
-                  const { uuid, name, description, arch, version, date, use_latest } = rowData;
-                  return (
+                {templateList.map(
+                  (
+                    {
+                      uuid,
+                      name,
+                      description,
+                      arch,
+                      version,
+                      date,
+                      use_latest,
+                      last_update_snapshot_error,
+                      last_update_task,
+                    }: TemplateItem,
+                    index,
+                  ) => (
                     <Tr key={uuid + index}>
                       <Td>
                         <Button
@@ -326,7 +338,11 @@ const TemplatesTable = () => {
                         </ConditionalTooltip>
                       </Td>
                       <Td>
-                        <StatusIcon rowData={rowData} />
+                        <StatusIcon
+                          uuid={uuid}
+                          last_update_snapshot_error={last_update_snapshot_error}
+                          last_update_task={last_update_task}
+                        />
                       </Td>
                       <Td>
                         <ConditionalTooltip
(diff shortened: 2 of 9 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-4670

README.md
```diff
@@ -36,7 +36,9 @@ Alternativly, simply add these lines to your /etc/hosts:
 
 4. `npm run build` (only required when setting up for the first time)
 
-5. `npm run start` to run against stage or prod environments. <br/>
+5. `npm run start:stage` to run against the stage environment <br/>
```
marusak, line 39: Thank you! However, shouldn't it be at least separate commit or maybe even PR? (Mostly interested here what is your git policy :) )

Andrewgdewar: I usually just add a separate section to my PR description for "Additional Changes" a separate commit is fine for this. 
I normally wouldn't bother with a separate PR, I find too tight of a commit policy results in me not fixing small bugs I otherwise would due to laziness.

xbhouse: thanks @Andrewgdewar, will clarify in the description and add in a separate commit for this!

src/Pages/Templates/TemplatesTable/TemplatesTable.tsx
```diff
@@ -250,22 +280,26 @@ const TemplatesTable = () => {
...
                 </Tr>
               </Thead>
               <Tbody>
-                {templateList.map(
-                  (
-                    { uuid, name, description, arch, version, date, use_latest }: TemplateItem,
-                    index,
-                  ) => (
+                {templateList.map((rowData: TemplateItem, index) => {
+                  const { uuid, name, description, arch, version, date, use_latest } = rowData;
```
Andrewgdewar, line 301: #nit: I think you could keep the simplified direct return here. 
No need to destructure from the variable.

src/Pages/Templates/TemplatesTable/components/StatusIcon.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+interface Props {
+  rowData: TemplateItem;
+}
+
+interface DescriptionProps {
+  error?: string;
+}
+
+const PopoverDescription = ({ error }: DescriptionProps) => <div>{error}</div>;
```
Andrewgdewar, line 26: This component likely isn't needed.

Andrewgdewar (approved):
A couple of small #nit comments, everything else looks good.

Andrewgdewar (approved):
A couple of small #nit comments, everything else looks good.

src/Pages/Templates/TemplatesTable/TemplatesTable.tsx
xbhouse: fix pushed

src/Pages/Templates/TemplatesTable/TemplatesTable.tsx
```diff
@@ -326,7 +338,11 @@ const TemplatesTable = () => {
                         </ConditionalTooltip>
                       </Td>
                       <Td>
-                        <StatusIcon rowData={rowData} />
```
Andrewgdewar: oh....
I get it, apologies.

src/Pages/Templates/TemplatesTable/TemplatesTable.tsx
Andrewgdewar: I see what you were doing there.. The change looks good, but what you had was likely fine. 😓

swadeley:
Hi

If you create the template before the included RHEL repos, e.g. Ansible, are introspected the template shows Invalid and never becomes Valid after the RHEL repo reaches Valid status. Is that out if scope?

xbhouse:
hi @swadeley, i think this is the same issue as the associated backend PR, because candlepin isn't in ephemeral and so the task that the status depends on doesn't run. i can adjust the logic here to check if we're in ephemeral so that the status isn't always Invalid, but we may only be able to test this fully in stage

swadeley:
Hi

nothing broken in ephemeral, merging to test further in stage.

PR #348, opened 2024-09-23
Build: Fixes a bug with arch/version display

Andrewgdewar:
## Summary

- Fixes a display bug with ArchVersion that would show "9" instead of "el9".


swadeley (approved):
ACK

PR #349, opened 2024-09-23
Build: Fix jest-preview version mismatch

Andrewgdewar:
## Summary

- Small follow-up fix to the previous node version update, fixes a small module version mismatch.


swadeley (approved):
ACK

PR #362, opened 2024-10-22
Fixes 4826: prevent adding systems if env not yet created

xbhouse:
## Summary

- Checks if environment is created (`rhsm_environment_created == true`) before allowing user to assign the template to a system
- Displays an error if the environment has not been created but the template's update task has finished
- (You will see some unintended behavior here when adding multiple templates until this [PR](https://github.com/content-services/content-sources-backend/pull/866) is merged)

## Testing steps

Since this depends on Patch's API, this needs to be tested on stage (`npm run start:stage`). You will also need to have [registered](https://github.com/content-services/content-sources-backend/blob/main/docs/register_client.md#registering-a-subscription-manager-client-to-stage-environment) at least one system on stage. This also cannot be tested in ephemeral as the task that sets the `rhsm_environment_created` flag depends on candlepin.

1. Create a template and select `Create template and add to systems` to navigate to the AddSystemModal
2. Select a system. The Assign button should be disabled for a few seconds and have a tooltip. Once the template's update task has finished and the environment was created, the button should be enabled

### DIFF
--- a/src/Pages/Templates/TemplateDetails/components/AddSystems/AddSystemModal.test.tsx
+++ b/src/Pages/Templates/TemplateDetails/components/AddSystems/AddSystemModal.test.tsx
@@ -2,7 +2,11 @@ import { render, waitFor } from '@testing-library/react';
 import AddSystemModal from './AddSystemModal';
 import { useQueryClient } from 'react-query';
 import { useSystemsListQuery } from 'services/Systems/SystemsQueries';
-import { defaultSystemsListItem } from 'testingHelpers';
+import {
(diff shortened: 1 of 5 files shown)


xbhouse:
seeing something weird here, it seems like `rhsm_environment_created` doesn't stay true if the template sits for a while? if we try to assign a template to a system a day or so after the template was created, its `rhsm_environment_created` value will be false after initially being set to true, so we won't be able to assign it unless we update the template. looking into it now

Andrewgdewar (approved):
Ack!

PR #387, opened 2024-11-21
"not (subscription or introspection)"

swadeley:
## Summary

Inhibiting subscription test because the fronted pr check runs in ephemeral which does not support real users and subscriptions. Also inhibiting introspection tests as that is not relevant to UI changes.

## Testing steps

tests pass


swadeley (commented):
ACK

dominikvagner (approved):
acked! 👏🏼

PR #392, opened 2024-12-03
HMS-5039: eliminates re-uploads of finished uploads

dominikvagner:
This PR adds a check before the upload of a file to see, if there is an artifact in pulp with the same hash and that's the case it skips the upload and sets the artifact href for it, which then can be used to add the file to the repo.

## Testing steps
1. Create 2 upload repositories.
2. Upload a big file to one of the repos (more than one chunk at least).
3. Let it upload and confirm the upload.
4. Try uploading the same file to the other repo and verify that the upload was skipped and the file can be added.


jlsherrill:
https://issues.redhat.com/browse/HMS-5039

jlsherrill:

PR #395, opened 2024-12-09
Build: Add open source badge

ochosi:
Add an 'open source' badge that let's anyone coming to the service easily discover the source code associated with it. Ideally we'd redirect to a page similar to the one Image Builder has that shows the architecture and has more docs. As a starting point, the frontend repository may be ok.

This is what it looks like:
image

Note: I hardcoded the page title to 'Repositories' - not sure if that's desirable.

### DIFF
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -1,5 +1,6 @@
 import { Text } from '@patternfly/react-core';
 import {
+  OpenSourceBadge,
   PageHeader as _PageHeader,
   PageHeaderTitle,
 } from '@redhat-cloud-services/frontend-components';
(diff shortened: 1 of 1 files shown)


src/components/Header/Header.tsx
```diff
@@ -32,12 +33,19 @@ interface Props {
   paragraph: string;
 }
 
-export default function Header({ title, ouiaId, paragraph }: Props) {
+export default function Header({ ouiaId, paragraph }: Props) {
```
jlsherrill, line 36: the 'title' is passed in from the repositories page and the templates page, so we probably want to continue that and just reference it via `{title}` below?

jlsherrill (approved):

PR #399, opened 2024-12-16
Build: Update dependabot

Andrewgdewar:
## Summary

This makes dependabot ignore major versions of a few modules, as it isn't capable of making the upgrade without assistance.


dominikvagner (approved):
makes sense 👍🏼 versions line up with the requirements in `package.json` (as the next major), approved! ✅

PR #403, opened 2024-12-23
HMS-5024: add use template modal

rverdile:
## Summary
Adds a "use template" button in the top left of the template details page that opens a modal with different sets of instructions on how to use templates.

image
image
image
image

### DIFF
--- a/src/Pages/Templates/TemplateDetails/TemplateDetails.test.tsx
+++ b/src/Pages/Templates/TemplateDetails/TemplateDetails.test.tsx
@@ -1,4 +1,4 @@
-import { render } from '@testing-library/react';
+import {fireEvent, render} from '@testing-library/react';
 import TemplateDetails from './TemplateDetails';
 import { defaultTemplateItem } from 'testingHelpers';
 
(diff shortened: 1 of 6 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-5024

jlsherrill:
This looks great, a couple of things to mention:

1.  We  probably should mention that `subscription-manager refresh` is needed after the steps which associates the system to the template
2. https://issues.redhat.com/browse/RHINENG-14351 is merged, so we can mention via the first method that you can use the UI to associate the system OR use this new api:
```
PATCH /template/$TEMPLATE_UUID/subscribed-systems  
```
but we could fill in the $TEMPLATE_UUID with the actual UUID.  I haven't tested this yet to confirm it works, i can do that on monday 

3.  for the curl and ansible method, we probably want to mention that system advisory information will not be available via the patch application

rverdile:
updated and updated the screenshots

jlsherrill:
I think there should be a space between these two buttons:
image

For the curl and ansible tabs, could you add an alert/warning box with something like:

```
Consuming a template in this way will result in the system not reporting applicable errata properly within Insights.  
```

jlsherrill:
everything else looks great!

jlsherrill (approved):

swadeley:
Hi

buttons have space between them:
image

tabs:
image


image

image

swadeley:
Hi @jlsherrill 

URLs are hardcoded I take it, no stage or ephemeral support then. OK?

PR #407, opened 2025-01-08
HMS-5197: Fixes List snapshot being disabled for read only users

Andrewgdewar:
## Summary

Small change to update permissions to allow read-only users to view snapshots

## Testing steps
- Get a read -only user, check before and after this change. 
- With this change they should now be permitted to look at the snapshotList within the kebab for each repository.


swadeley:
Hi

lgtm

Screenshot at 2025-01-09 14-46-36

swadeley (approved):
ACK

PR #412, opened 2025-01-16
HMS-5267: template wizard date picker fix, UX changes

dominikvagner:
## Summary
This PR fixes the existing problems with the date picker (weird input behavior, problematic validation and it makes the "next" action dependent on the input) in our template wizard and also implements proposed UX changes.

image

## Testing steps
1. Verify the date picker works correctly/better than before. (i.e. no "Invalid Date" strings, you can type out the date normally)
2. Check that the warning and the listed repos are showing up correctly. (i.e. all repos that have all snapshots past the selected date are showing up only when they should)
3. Make sure the "Next" button doesn't allow you to go further if the "Use up to specific date" option is selected and the date isn't valid.

### DIFF
--- a/src/Pages/Templates/TemplatesTable/components/AddTemplate/steps/SetUpDateStep.tsx
+++ b/src/Pages/Templates/TemplatesTable/components/AddTemplate/steps/SetUpDateStep.tsx
@@ -47,7 +47,7 @@ export default function SetUpDateStep() {
   useEffect(() => {
     setTemplateRequest({
       ...templateRequest,
-      date: templateRequest ? formatDateForPicker(templateRequest.date) : '',
+      date: templateRequest.date ? formatDateForPicker(templateRequest.date) : '',
     });
   }, []);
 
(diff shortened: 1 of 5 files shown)


src/Pages/Templates/TemplatesTable/components/AddTemplate/steps/SetUpDateStep.tsx
```diff
@@ -128,122 +123,58 @@ export default function SetUpDateStep() {
...
           onChange={() => {
             if (templateRequest.use_latest) {
               setTemplateRequest((prev) => ({ ...prev, use_latest: false, date: '' }));
             }
           }}
         />
+        <DatePicker
+          value={templateRequest.date ?? ''}
+          required
+          requiredDateOptions={{ isRequired: true }}
```
xbhouse, line 138: i notice if i select `Use up to specific date` and click inside the date picker and enter an invalid date (or leave it empty) then instead decide to select `Use latest content`, the `Date cannot be blank` error remains but I can proceed in the wizard. i'd expect the error about the invalid date to disappear if `Use latest content` is selected. 

i wonder if this is because a date is set to be required here?

swadeley:
lgtm

image

swadeley:
I confirm invalid date (2022-10-33) keeps "Next" button disabled.

xbhouse (approved):
nice job, looks great! ack! 🎉

PR #413, opened 2025-01-17
Fixes kebab being disabled on content list page

Andrewgdewar:
## Summary

Kebab no longer needs to be disabled on the content list page. 

This fix is the result of a failing test! Huzzah!


mayurilahane:
verified that while snapshot is in preogress the kebab is enabled as expected 

image

jlsherrill (approved):

xbhouse (approved):

PR #423, opened 2025-02-06
Tests: Don't run API tests in frontend tests

marusak:
There is not any value in doing this. It makes these tests slower, more complex and opens it up to being flaky or broken. Let's keep it simple and only do things, that are bringing real value.


Andrewgdewar (approved):

marusak:
Test failures are unrelated, I will merge. I want to debug those Upload file ones, but I want to first get in better reporter :) And this is prerequisite

PR #424, opened 2025-02-06
HMS-5426: add PW test for RH repos

xbhouse:
## Summary

- Depends on this [PR](https://github.com/content-services/content-sources-frontend/pull/440) adding in the CDN cert
- Adds test to validate the imported RH repo exists, has a status of "Valid", and has at least one snapshot

## Testing steps

- Tests pass, all criteria in linked issue is covered

### DIFF
--- a/_playwright-tests/UI/RedHatRepo.spec.ts
+++ b/_playwright-tests/UI/RedHatRepo.spec.ts
@@ -1,0 +1,55 @@
+import { test, expect } from '@playwright/test';
+import { navigateToRepositories } from './helpers/navHelpers';
+import { closePopupsIfExist } from './helpers/helpers';
+
+test.describe('Red Hat Repositories', () => {
+    const smallRHRepo = 'Red Hat Ansible Engine 2 for RHEL 8 x86_64 (RPMs)'
+
+    test.beforeEach(async ({ page }) => {
+        await test.step('Navigate to repositories page', async () => {
+            await navigateToRepositories(page);
+            await closePopupsIfExist(page);
+        });
+        await test.step('Navigate to Red Hat repositories', async () => {
+            await page.getByRole('button', { name: 'Red Hat' }).click();
+        });
+    });
+
+    test('Wait for repository status to be "Valid"', async ({ page }) => {
+        test.setTimeout(600000);
+        const status = page
+            .getByRole('row')
+            .filter({ hasText: smallRHRepo })
+            .getByRole('gridcell')
+            .filter({ hasText: 'Valid' });
+        
+        await test.step('Filter repos by name', async () => {
+            await page.getByPlaceholder('Filter by name/url').fill(smallRHRepo);
+        });
+
+        await test.step('Wait for a valid status', async () => {
+            await status.waitFor({ state: 'visible' });
+        });
+    });
+
+    test('Check repository snapshots', async ({ page }) => {
+        await test.step('Open the snapshots list modal', async () => {
+            await page
+                .getByRole('row')
+                .filter({ hasText: smallRHRepo })
+                .getByLabel('Kebab toggle')
+                .click();
+            await expect(page.getByRole('menuitem', { name: 'View all snapshots' })).toBeVisible();
+            await page.getByRole('menuitem', { name: 'View all snapshots' }).click();
+            await expect(page
+                .getByTestId('snapshot_list_modal')
+                .filter({ hasText: smallRHRepo }))
+                .toBeVisible();
+        });
+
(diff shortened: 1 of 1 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-5426

_playwright-tests/UI/RedHatRepo.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+import { test, expect } from '@playwright/test';
+import { navigateToRepositories } from './helpers/navHelpers';
+import { closePopupsIfExist } from './helpers/helpers';
+
+describe('Red Hat Repositories', () => {
+  test('Red Hat repository exists and is valid', async ({ page }) => {
+    await navigateToRepositories(page);
+    await closePopupsIfExist(page);
+    const redHatToggle = page.getByTestId('redhat-repositories-toggle');
+    await redHatToggle.click();
```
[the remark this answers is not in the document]
marusak: Also TestId is not descriptive enough :man_shrugging: I could not understand what this is targetting by looking at it. Either better selector or comment would be appreciated :) 

`await page.getByRole('button', { name: 'Red Hat', exact: true }).click();` should work though

_playwright-tests/UI/RedHatRepo.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+
+    await expect(page.getByText('Red Hat Ansible Engine 2 for RHEL 8 x86_64 (RPMs)')).toBeVisible();
+    await expect(page.getByText('Valid')).toBeVisible();
+    await expect(page.getByText('Last snapshot')).toBeVisible();
```
marusak, line 15: These two are too generic. They only check that any repo in the list has `Valid` and `Last snapshot` visible. It would fail if there would be more than one repo (because it is enforced that only one element is being targeted).

I guess in our local setup we have only this one repo, so it works. But if we would run this test against stage it would not work. I think for writing cleaner test it would be better to target one specific repo and check on it, that it contains these values. 

Storing the first selector and then filtering might be easy and readable :) https://playwright.dev/docs/locators#filtering-locators

xbhouse: oh yep i meant to add this in! just pushed it up in its current state yesterday to test the gh action :)

_playwright-tests/UI/RedHatRepo.spec.ts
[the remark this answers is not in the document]
xbhouse: ah yeah using the button selector did work, it wasn't before (i thought because of the span inside the button). i can use this instead here, thanks!

playwright.config.ts
```diff
@@ -20,7 +20,7 @@ export default defineConfig({
         ['html', { outputFolder: 'playwright-report' }],
       ]
     : 'list',
-  timeout: process.env.CI ? 60000 : 30000,
+  timeout: process.env.CI ? 300000 : 30000,
```
marusak, line 23: Why should we change this globally? This will make failing too slow - e.g. if you have failing test, you will be waiting for results significantly longer.

xbhouse: overriding that locally like this `await status.waitFor({ state: 'visible', timeout: 300000 });` didn't seem to work, let me try test.setTimeout

_playwright-tests/UI/RedHatRepo.spec.ts
```diff
@@ -0,0 +1,57 @@
+import { describe } from 'node:test';
+import { test, expect } from '@playwright/test';
+import { navigateToRepositories } from './helpers/navHelpers';
+import { closePopupsIfExist } from './helpers/helpers';
+
+describe('Red Hat Repositories', () => {
```
dominikvagner, line 6: it would be better to change this to the describe method from playwright test (`test.describe()`) like in the other suites, that one works better with playwright adapters 😅

xbhouse: oof yes missed changing this 😆 thank you!

playwright.config.ts
Andrewgdewar: You can adjust the timeout for your specific test by passing in the value to the expect method.
expect (something).doesAThing({timeout:1234})

Andrewgdewar:
The latest test flakiness thing I have seen when running the front-end locally. Sometimes you have to refresh, things just don't load...

Andrewgdewar (approved):
Looks good!

swadeley (approved):
ACK

PR #425, opened 2025-02-06
Test: Popular Repos

swadeley:
[HMS-5424] UI - Playwright POC - Popular repos

## Summary

Test navigating to, and adding, Popular repos

### DIFF
--- a/_playwright-tests/UI/PopularRepo.spec.ts
+++ b/_playwright-tests/UI/PopularRepo.spec.ts
@@ -1,0 +1,66 @@
+import { describe } from 'node:test';
+import { test, type Page } from '@playwright/test';
+import { navigateToRepositories } from './helpers/navHelpers';
+import { deleteAllRepos } from './helpers/deleteRepositories';
+import { closePopupsIfExist } from './helpers/helpers';
+
+describe('Popular Repositories', () => {
+  test('Clean - Delete any current repos that exist', async ({ page }) => {
+    await deleteAllRepos(page);
+  });
+
+  test('Navigate to Popular Repos tab, assert the Add button is displayed', async ({ page }) => {
+    await navigateToRepositories(page);
+    await closePopupsIfExist(page);
+
+    const AddMenuButton = page.locator('[data-ouia-component-id="add_checked_repos-with-snap"]');
+    await page.getByRole('link', { name: 'Popular repositories' }).click();
+
+
+    // Wait for the Add button to become enabled (up to 10 seconds)
+    await AddMenuButton.first().isEnabled({ timeout: 10000 });
+  });
+
+  test('Add popular repos', async ({ page }) => {
+    await navigateToRepositories(page);
+    await closePopupsIfExist(page);
+
+    const AddMenuButton = page.locator('[data-ouia-component-id="add_popular_repo_toggle"]');
+    const AddButton = page.locator('[data-ouia-component-id="add_popular_repo"]');
+    const AddButtonNoSnap = page.locator('[data-ouia-component-id="add-popular_repo_without-snapshotting"]');
+
+    await page.goto('https://stage.foo.redhat.com:1337/insights/content/repositories#SIDs=&tags=');
+    await page.getByRole('link', { name: 'Popular repositories' }).click();
+
+    // Wait for the Add button to become enabled (up to 10 seconds)
+    await AddButton.first().isEnabled({ timeout: 10000 });
+    await page.getByRole('row', { name: 'Select row 0 EPEL 9' }).getByLabel('Select', { exact: true }).click();
+    await page.getByRole('menuitem', { name: 'Add without snapshotting' }).click();
+    await page.getByRole('gridcell', { name: 'Add Select' }).getByLabel('Select').click();
+    await page.getByRole('menuitem', { name: 'Add without snapshotting' }).click();
+    await page.getByRole('link', { name: 'Your repositories' }).click();
+    await page.getByRole('textbox', { name: 'Filter by name/url' }).click();
+    await page.getByRole('textbox', { name: 'Filter by name/url' }).fill('EPEL');
+    await page.locator('#topPaginationWidgetId-top-toggle').click({
+      button: 'right'
+    });
+    await page.locator('#topPaginationWidgetId-top-toggle').click();
+    await page.getByRole('menuitem', { name: '10 per page' }).press('Escape');
+    await page.getByRole('checkbox', { name: 'Select row 0' }).check();
+    await page.getByRole('button', { name: 'Actions' }).click();
+    await page.getByRole('menuitem', { name: 'Remove 1 repositories' }).click();
+    await page.getByRole('button', { name: 'Remove' }).click();
+    await page.locator('#topPaginationWidgetId-top-toggle').click();
+    await page.getByRole('menuitem', { name: '10 per page' }).press('Escape');
+    await page.getByRole('checkbox', { name: 'Select all rows' }).check();
+    await page.getByRole('button', { name: 'Actions' }).click();
+    await page.getByRole('menuitem', { name: 'Remove 1 repositories' }).click();
+    await page.getByRole('button', { name: 'Remove' }).click();
+    await page.getByRole('link', { name: 'Popular repositories' }).click();
+    await page.getByRole('checkbox', { name: 'Select all rows' }).check();
+    await page.locator('#topPaginationWidgetId-top-toggle').click();
+    await page.getByRole('menuitem', { name: '10 per page' }).press('Escape');
+
+
+  });
+});


_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -0,0 +1,66 @@
+import { describe } from 'node:test';
+import { test, type Page } from '@playwright/test';
+import { navigateToRepositories } from './helpers/navHelpers';
+import { deleteAllRepos } from './helpers/deleteRepositories';
+import { closePopupsIfExist } from './helpers/helpers';
+
+describe('Popular Repositories', () => {
+  test('Clean - Delete any current repos that exist', async ({ page }) => {
```
marusak, line 8: I am not sure if this is a good practice to create test cases for prerequistes. It is overhead for each test case - e.g. starting browser, verifying session, navigating to console dot, loading all of it's resources, navigation to our app...

I think we should not create test cases unnecessarily. This could be easily on top of your test - if it fail your test fails as well and you will see where. Or is there any other reason why is it done separately?

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    await deleteAllRepos(page);
+  });
+
+  test('Navigate to Popular Repos tab, assert the Add button is displayed', async ({ page }) => {
```
marusak, line 12: Similarly here. It is rather simple test - does it need to be separate?

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    await navigateToRepositories(page);
+    await closePopupsIfExist(page);
+
+    const AddMenuButton = page.locator('[data-ouia-component-id="add_checked_repos-with-snap"]');
```
marusak, line 16: For this you can use `getByTestId` - see https://github.com/content-services/content-sources-frontend/pull/424/commits/d3b785345ce84f4d65860346581829a97fc5bc47

Maybe we should add this commit separately to enable others to use it

marusak: Also, wouldn't there by selector that describes it better? E.g. with text or so? Now it is not clear (to me :) ) what button it actually targets.

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    await page.getByRole('link', { name: 'Popular repositories' }).click();
+
+
+    // Wait for the Add button to become enabled (up to 10 seconds)
+    await AddMenuButton.first().isEnabled({ timeout: 10000 });
```
marusak, line 21: Why is there `.first()`?

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  test('Add popular repos', async ({ page }) => {
+    await navigateToRepositories(page);
+    await closePopupsIfExist(page);
+
+    const AddMenuButton = page.locator('[data-ouia-component-id="add_popular_repo_toggle"]');
```
marusak, line 28: This does not seem to be used? :thinking:

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    const AddButton = page.locator('[data-ouia-component-id="add_popular_repo"]');
+    const AddButtonNoSnap = page.locator('[data-ouia-component-id="add-popular_repo_without-snapshotting"]');
```
marusak, line 30: This does not seem to be used either

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    // Wait for the Add button to become enabled (up to 10 seconds)
+    await AddMenuButton.first().isEnabled({ timeout: 10000 });
+  });
+
+  test('Add popular repos', async ({ page }) => {
+    await navigateToRepositories(page);
+    await closePopupsIfExist(page);
+
+    const AddMenuButton = page.locator('[data-ouia-component-id="add_popular_repo_toggle"]');
+    const AddButton = page.locator('[data-ouia-component-id="add_popular_repo"]');
```
marusak, line 29: This is used exactly once - maybe just use it inline?

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    const AddButtonNoSnap = page.locator('[data-ouia-component-id="add-popular_repo_without-snapshotting"]');
+
+    await page.goto('https://stage.foo.redhat.com:1337/insights/content/repositories#SIDs=&tags=');
```
marusak, line 32: Don't navigate like this. We have BASE_URL set and we should respect that. So either just https://playwright.dev/docs/api/class-page#page-goto or Andrew actually created helper function to navigate to repos. (that you already used in this test).

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    await page.getByRole('link', { name: 'Popular repositories' }).click();
+
+    // Wait for the Add button to become enabled (up to 10 seconds)
+    await AddButton.first().isEnabled({ timeout: 10000 });
+    await page.getByRole('row', { name: 'Select row 0 EPEL 9' }).getByLabel('Select', { exact: true }).click();
+    await page.getByRole('menuitem', { name: 'Add without snapshotting' }).click();
+    await page.getByRole('gridcell', { name: 'Add Select' }).getByLabel('Select').click();
+    await page.getByRole('menuitem', { name: 'Add without snapshotting' }).click();
+    await page.getByRole('link', { name: 'Your repositories' }).click();
+    await page.getByRole('textbox', { name: 'Filter by name/url' }).click();
```
marusak, line 42: Do we need this? I already asked Andrew about it. If we do, than we actually may have bug in our code :)

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    await page.getByRole('textbox', { name: 'Filter by name/url' }).fill('EPEL');
+    await page.locator('#topPaginationWidgetId-top-toggle').click({
+      button: 'right'
+    });
+    await page.locator('#topPaginationWidgetId-top-toggle').click();
```
marusak, line 47: Some comment what is happening now would be nice :)

marusak: Also why this step? You send escape to it

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    await page.getByRole('menuitem', { name: '10 per page' }).press('Escape');
+    await page.getByRole('checkbox', { name: 'Select row 0' }).check();
+    await page.getByRole('button', { name: 'Actions' }).click();
+    await page.getByRole('menuitem', { name: 'Remove 1 repositories' }).click();
+    await page.getByRole('button', { name: 'Remove' }).click();
+    await page.locator('#topPaginationWidgetId-top-toggle').click();
```
marusak, line 53: Is this some force to get updates?

marusak (changes requested):
Nice! Couple of comments and question, but I like where this is heading :)

_playwright-tests/UI/PopularRepo.spec.ts
swadeley: Hi, just trying to lean how multiple "tests" fit within a "describe" block. Will change that. Thank you.

_playwright-tests/UI/PopularRepo.spec.ts
swadeley: No, just learning

_playwright-tests/UI/PopularRepo.spec.ts
swadeley: oh yes, will fix

_playwright-tests/UI/PopularRepo.spec.ts
swadeley: This is partly habit, we currently define buttons separately from the test. I would prefer the test to be consistent (either define above or inline rather than mix methods).

_playwright-tests/UI/PopularRepo.spec.ts
swadeley: escape bad copy paste from auto generated code

_playwright-tests/UI/PopularRepo.spec.ts
jlsherrill: with https://github.com/content-services/content-sources-frontend/pull/428  you can use  page.getByTestId('add_popular_repo')    lets get that merged!

xbhouse: just merged so that's an option, you'll just need to rebase

_playwright-tests/UI/PopularRepo.spec.ts
Andrewgdewar: This may be a good spot to try the "test.step" method that @dominikvagner  introduced.

_playwright-tests/UI/PopularRepo.spec.ts
swadeley: will remove

_playwright-tests/UI/PopularRepo.spec.ts
swadeley: From from auto generated code, refactoring

_playwright-tests/UI/PopularRepo.spec.ts
swadeley: from auto generated code, removing

_playwright-tests/UI/PopularRepo.spec.ts
swadeley: Ok, now using `test.step`

_playwright-tests/UI/PopularRepo.spec.ts
swadeley: Done that, thank you

_playwright-tests/UI/PopularRepo.spec.ts
swadeley: I had to revert that use of `test.step`, test looses session and then cannot find repos to remove.

dominikvagner: `test.step()` should not impact the session, playwright uses one for the whole test. 
it's use is for adding a step to the test report and for us "documenting" what the test does in more detail, so it most likely wasn't caused by the step func 💭😅

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+    // Open the drop-down menu
+    await page.getByTestId('add-selected-dropdown-toggle-no-snap').click();
+    // Click the first and only button in the drop-down menu
+    await page.getByTestId('add-selected-repos-no-snap').click();
+    // Multiple toast pop-ups about snapshotting and the code tries to close that
+    // Move to Custom repo tab
+    await page.getByRole('link', { name: 'Your repositories' }).click();
+    await expect(page.getByTestId('custom_repositories_table')).toBeVisible();
+
+    // Filter for EPEL and pick the first row
```
marusak, line 41: All of these could be `steps` and having it commented in that way

_playwright-tests/UI/helpers/deletePopularRepositories.ts
```diff
@@ -0,0 +1,36 @@
+import { expect, type Page } from '@playwright/test';
+
+export const deleteAllPopularRepos = async ({ request }: Page) => {
+  const response = await request.get('/api/content-sources/v1/popular_repositories/?');
+
+  // Ensure the request was successful
+  console.log('response.status');
```
marusak, line 7: Forgotten debug?

_playwright-tests/UI/helpers/deletePopularRepositories.ts
```diff
@@ -0,0 +1,36 @@
...
+        data: {
+          uuids: uuidList,
+        },
+      });
+
+      // Ensure the deletion was successful
+      expect(result.status()).toBe(204);
+    } catch (error) {
+      console.error('Failed to delete repositories:', error);
+      throw error; // Optionally re-throw the error if you need to fail the test
```
marusak, line 31: Optionally? :) I guess you can drop the comment now

marusak (changes requested):
Looks amazing! I think you could use more `steps` and less comments (replace comments by steps). Otherwise just nitpicks :)

Can you also please rebase to get new run also with reporter? Thanks!

_playwright-tests/UI/helpers/deletePopularRepositories.ts
swadeley: yes, removing

_playwright-tests/UI/helpers/deletePopularRepositories.ts
swadeley: removing

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+    // Select the first row then use main side toggle menu to add it
+    await page
+      .getByRole('row', { name: 'EPEL 9 Everything x86_64' })
+      .getByLabel('Select row 0', { exact: true })
+      .click();
+    // Open the drop-down menu
+    await page.getByTestId('add-selected-dropdown-toggle-no-snap').click();
+    // Click the first and only button in the drop-down menu
+    await page.getByTestId('add-selected-repos-no-snap').click();
+    // Multiple toast pop-ups about snapshotting and the code tries to close that
```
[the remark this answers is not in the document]
swadeley: Updated

_playwright-tests/UI/PopularRepo.spec.ts
swadeley: Done that now

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    // Select the first row then use main side toggle menu to add it
+    await page
+      .getByRole('row', { name: 'EPEL 9 Everything x86_64' })
+      .getByLabel('Select row 0', { exact: true })
+      .click();
+    // Open the drop-down menu
+    await page.getByTestId('add-selected-dropdown-toggle-no-snap').click();
+    // Click the first and only button in the drop-down menu
+    await page.getByTestId('add-selected-repos-no-snap').click();
```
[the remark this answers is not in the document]
swadeley: That button is just a done arrow, no text

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+      await expect(page.getByTestId('popular_repos_table')).toBeVisible();
+    });
+
+    await test.step('Select and add EPEL 9', async () => {
+      await page
+        .getByRole('row', { name: 'EPEL 9 Everything x86_64' })
+        .getByLabel('Select row 0', { exact: true })
+        .click();
+      await page.getByTestId('add-selected-dropdown-toggle-no-snap').click();
+      await page.getByTestId('add-selected-repos-no-snap').click();
```
marusak, line 26: If this would be `await page.getByRole("button", {name: "Add 1 repository without snapshotting" }).click()` it would both read better but also would test that we show the text correctly (e.g. it shows `1` and then we could have another test that would test it shows `2`).

Also the test could first verify that the button is disabled, then it is enabled with one and shows `1` and with two shows `2`...

But that is good as a followup :)

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+   });
+
+    await test.step('Select and add EPEL 8', async () => {
```
marusak, line 29: Why doing it in two steps? and not select both?

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+      await page
+        .getByRole('row', { name: 'EPEL 8 Everything x86_64' })
+        .getByLabel('Select row 1', { exact: true })
+        .click();
+      await page.getByTestId('add-selected-dropdown-toggle-no-snap').click();
+      await page.getByTestId('add-selected-repos-no-snap').click();
+    });
```
marusak, line 36: After adding them it would be great to check that those lines actually contain `Remove` button instead of `Add`.

marusak (approved):
Looking good! I see that you prefer using TestID :D 
The test could be more detailed and verify more cases (e.g. adding two repos, verifying button is disabled when no repo is selected, checking that we show the correct number when multiple repos are selected...) but all of those are great for followup :) 

Thanks!

_playwright-tests/UI/PopularRepo.spec.ts
swadeley: No special reason , I will change it

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+    await deleteAllPopularRepos(page);
+
+    await navigateToRepositories(page);
+    await closePopupsIfExist(page);
+    await expect(page).toHaveTitle('Repositories - Content | RHEL');
+
+    await test.step('Select the Popular repos tab', async () => {
+      await page.getByRole('link', { name: 'Popular repositories' }).click();
+      await expect(page.getByTestId('popular_repos_table')).toBeVisible();
+      await page.getByRole('button', { name: 'Add selected repositories' });
```
marusak, line 18: Is this enough? Don't you need to assert something? E.g. "toBeVisible()` or do something e.g. `click()`?

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+    });
+
+    await test.step('Select and add EPEL 9', async () => {
+      await page
+        .getByRole('row', { name: 'EPEL 9 Everything x86_64' })
+        .getByLabel('Select row 0', { exact: true })
+        .click();
+      await page.getByTestId('add-selected-dropdown-toggle-no-snap').click();
+      await page.getByRole('menuitem', { name: 'Add 1 repositories without snapshotting' });
```
marusak, line 27: Same, shouldn't you click on it? :thinking:

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+      await page
+        .getByRole('row', { name: 'EPEL 9 Everything x86_64' })
+        .getByTestId('remove_popular_repo')
+        .getByText('Remove');
```
marusak, line 31: shouldn;t you expect it to be visible or something?

Or is this selector enough and it waits for it to be present?

_playwright-tests/UI/PopularRepo.spec.ts
swadeley: I have updated that section to click the menu button based on its name and role.

_playwright-tests/UI/PopularRepo.spec.ts
swadeley: Updated the at section

_playwright-tests/UI/PopularRepo.spec.ts
swadeley: I have added a check for that text

_playwright-tests/UI/PopularRepo.spec.ts
swadeley: We do not need to click as that is the expected default state, I can change that last line to use "expect"

marusak: right, when we just need to have something visible, then `toBeVisible()` should be used.

_playwright-tests/UI/PopularRepo.spec.ts
swadeley: I changed that section, no just click the second time so we can check the text changes (1 Repo, 2 Repos)

Andrewgdewar:
Failing tests due to flaky introspect tests.

PR #426, opened 2025-02-07
Test: Introduce playwright reporter

marusak:
It is built on top of #423 so disregard the first commit. I want to land that first as it causes issues ATM and this PR might need some iterations :) 

This reporter does following:
    - Publish html tests results to github pages
    - Publish comment with summary and url link to test results
    - When posting comments, previous comment is deleted
    
Also introducing github action to delete all tests results older than 7 days.

### DIFF
--- a/.github/workflows/playwright-actions.yml
+++ b/.github/workflows/playwright-actions.yml
@@ -191,11 +191,11 @@ jobs:
           .github/workflowScripts/waitForBackend.sh
 
       - name: Run front-end Playwright tests
-        run: yarn playwright test > pw-output
+        run: yarn playwright test | tee pw-output
 
       - name: Store test summary
         if: ${{ !cancelled() }}
-        run: tail -1 pw-output > playwright-report/summary
+        run: tail -2 pw-output | head -1 > playwright-report/summary
 
       - name: Publish front-end Test Report
         uses: ctrf-io/github-test-reporter@v1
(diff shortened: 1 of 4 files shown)


marusak:
I believe this works. According to the [internet](https://stackoverflow.com/questions/74957218/what-is-the-difference-between-pull-request-and-pull-request-target-event-in-git) we need `on: pull_request_target`.

- I tested that the workflow works in https://github.com/content-services/content-sources-frontend/pull/434#issuecomment-2649125204
- I tested that if it is merged it works as well in https://github.com/marusak/pw-test/pull/4 (Dominik sent PR from his fork and the same workflow triggered)
- `on:pull_request_target` however needs this to land first. Are we willing to try it? :) 

If we merge it and it fails, I am buying everyone beer/cake for every fix I need to send :D

Andrewgdewar:
> I believe this works. According to the [internet](https://stackoverflow.com/questions/74957218/what-is-the-difference-between-pull-request-and-pull-request-target-event-in-git) we need `on: pull_request_target`.
> 
> * I tested that the workflow works in [Test: PW reporter #434 (comment)](https://github.com/content-services/content-sources-frontend/pull/434#issuecomment-2649125204)
> * I tested that if it is merged it works as well in [style: remove extra line marusak/pw-test#4](https://github.com/marusak/pw-test/pull/4) (Dominik sent PR from his fork and the same workflow triggered)
> * `on:pull_request_target` however needs this to land first. Are we willing to try it? :)
> 
> If we merge it and it fails, I am buying everyone beer/cake for every fix I need to send :D

pull_request_target Does not worth with private runners and codebuild. 
Either run your automation after this and grab the created artifact, or find another way 😢

PR #430, opened 2025-02-10
Test: fix describe usage, use playwright test describe

dominikvagner:
## Summary
This PR is a small fix that changes the so far used describe method for test suites to one that comes from `playwright` rather then from `node:test`. This will enable integration with playwright adapters and minimize imports.

## Testing steps
- Verify that tests run as expected.
- If you are using playwright through some sort of plugin/extension, you should now have option to run the whole describe with one action. (i.e.: if you are using VS Code, you will have a play/run button next to the describe 🎉 )


marusak (approved):
nice, thanks!

PR #432, opened 2025-02-10
HMS-5425: add PW tests for introspect repos

dominikvagner:
## Summary
This PR adds PW tests for introspect only repositories. Also adds new helper `getRowByName()` that simplifies getting a locator for a specific row. Small change to `deleteAllRepos()` helper to minimize repeating ourselves. 

## Testing steps
- Verify that all tests in `IntrospectRepo.spec.ts` pass and test the necessary stuff listed in the related issue description.

### DIFF
--- a/_playwright-tests/UI/IntrospectRepo.spec.ts
+++ b/_playwright-tests/UI/IntrospectRepo.spec.ts
@@ -1,0 +1,85 @@
+import { test, expect } from '@playwright/test';
+import { navigateToRepositories } from './helpers/navHelpers';
+import { closePopupsIfExist, getRowByName } from './helpers/helpers';
+import { deleteAllRepos } from './helpers/deleteRepositories';
+
+test.describe('Introspect Repositories', () => {
+  const repoName = 'introspection-test';
+  const repoUrl = 'https://fedorapeople.org/groups/katello/fakerepos/zoo/';
+
+  test.beforeEach(async ({ page }) => {
+    await test.step('Navigate to the repository page', async () => {
+      await navigateToRepositories(page);
+      await closePopupsIfExist(page);
+    });
+  });
+
+  test('Create an introspection repository', async ({ page }) => {
+    await test.step('Cleanup repository, if using the same url', async () => {
+      await deleteAllRepos(page, `&url=${repoUrl}`);
+    });
+
+    await test.step('Open the add repository modal', async () => {
+      await page.getByRole('button', { name: 'Add repositories' }).first().click();
+      await expect(page.getByText('Add custom repositories')).toBeVisible();
+    });
+
+    await test.step('Fill the create repository form', async () => {
+      await page.getByPlaceholder('Enter name').fill(repoName);
+      await page.getByPlaceholder('https://').fill(repoUrl);
+    });
+
+    await test.step('Submit the form and wait for modal to disappear', async () => {
+      await Promise.all([
+        page.getByRole('button', { name: 'Save' }).click(),
+        page.waitForResponse(
+          (resp) =>
+            resp.url().includes('/bulk_create/') && resp.status() >= 200 && resp.status() < 300,
+        ),
+        expect(page.getByText('Add custom repositories')).not.toBeVisible(),
+      ]);
+    });
+  });
+
+  test('Wait for valid introspection status', async ({ page }) => {
+    await test.step('Wait for status to be "Valid"', async () => {
+      const row = await getRowByName(page, repoName);
+      await expect(row.getByText('Valid')).toBeVisible();
+    });
+  });
+
+  test('Check introspected repository packages', async ({ page }) => {
+    await test.step('Open the packages modal', async () => {
+      const row = await getRowByName(page, repoName);
+      await row.getByRole('gridcell', { name: '8', exact: true }).locator('a').click();
+      await expect(page.getByText('View list of packages')).toBeVisible();
+    });
+
+    await test.step('Check the modal for expected content', async () => {
+      await Promise.all([
+        expect(page.getByText('noarch')).toHaveCount(8),
+        expect(page.getByText('cheetah')).toBeVisible(),
+        expect(page.getByText('0.3').first()).toBeVisible(),
+        expect(page.getByText('0.8').first()).toBeVisible(),
+        expect(page.getByText('noarch').first()).toBeVisible(),
+      ]);
+    });
+  });
+
+  test('Delete introspected repository', async ({ page }) => {
+    await test.step('Delete created repository', async () => {
+      const row = await getRowByName(page, repoName);
+      await row.getByLabel('Kebab toggle').first().click();
+      await row.getByRole('menuitem', { name: 'Delete' }).click();
+      await expect(page.getByText('Remove repositories?')).toBeVisible();
+
+      await Promise.all([
+        page.waitForResponse(
+          (resp) =>
+            resp.url().includes('bulk_delete') && resp.status() >= 200 && resp.status() < 300,
+        ),
+        page.getByRole('button', { name: 'Remove' }).click(),
+      ]);
+    });
+  });
+});
--- a/_playwright-tests/UI/IntrospectRepo.spec.ts
+++ b/_playwright-tests/UI/IntrospectRepo.spec.ts
@@ -56,6 +56,7 @@ test.describe('Introspect Repositories', () => {
       const row = await getRowByName(page, repoName);
       await row.getByRole('gridcell', { name: repoPackageCount, exact: true }).locator('a').click();
       await expect(page.getByText('View list of packages')).toBeVisible();
+      await expect(page.getByText('Version').first()).toBeVisible();
     });
 
     await test.step('Check the modal for expected content', async () => {
--- a/_playwright-tests/UI/IntrospectRepo.spec.ts
+++ b/_playwright-tests/UI/IntrospectRepo.spec.ts
@@ -10,6 +10,7 @@ test.describe('Introspect Repositories', () => {
   const repoVersion = '0.3';
   const repoRelease = '0.8';
   const repoArch = 'noarch';
+  const testPackage = 'cheetah';
 
   test.beforeEach(async ({ page }) => {
     await test.step('Navigate to the repository page', async () => {
@@ -60,12 +61,13 @@ test.describe('Introspect Repositories', () => {
     });
 
     await test.step('Check the modal for expected content', async () => {
+      const row = page.getByRole('row').filter({ has: page.getByText(testPackage) });
       await Promise.all([
         expect(page.getByText(repoArch)).toHaveCount(8),
-        expect(page.getByText('cheetah')).toBeVisible(),
-        expect(page.getByText(repoVersion).first()).toBeVisible(),
-        expect(page.getByText(repoRelease).first()).toBeVisible(),
-        expect(page.getByText(repoArch).first()).toBeVisible(),
+        expect(row.getByText(testPackage)).toBeVisible(),
+        expect(row.getByText(repoVersion)).toBeVisible(),
+        expect(row.getByText(repoRelease)).toBeVisible(),
+        expect(row.getByText(repoArch)).toBeVisible(),
       ]);
     });
   });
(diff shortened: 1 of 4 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-5425

_playwright-tests/UI/IntrospectRepo.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    });
+  });
+
+  test('Create an introspection repository', async ({ page }) => {
+    await test.step('Cleanup repository, if using the same url', async () => {
+      await deleteAllRepos(page, `&url=${repoUrl}`);
+    });
+
+    await test.step('Open the add repository modal', async () => {
+      await page.getByRole('button', { name: 'Add repositories' }).first().click();
```
marusak, line 23: Why does this need the `first()` selector?

_playwright-tests/UI/IntrospectRepo.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+        page.waitForResponse(
+          (resp) =>
+            resp.url().includes('/bulk_create/') && resp.status() >= 200 && resp.status() < 300,
+        ),
+        expect(page.getByText('Add custom repositories')).not.toBeVisible(),
+      ]);
+    });
+  });
+
+  test('Wait for valid introspection status', async ({ page }) => {
```
marusak, line 44: So this `test` continues in the same browser as previous `test`? Or it kills the browser and starts a new session?

marusak (changes requested):
Looks amazing! I have couple questions, some of which might be just my lack of  understanding :)

_playwright-tests/UI/IntrospectRepo.spec.ts
dominikvagner: if the empty state on the repositories page would be displayed, then there would be two buttons with "Add repositories" and without the `first()` the test would fail on violating the strict locator rule 😅

_playwright-tests/UI/helpers/helpers.ts
```diff
@@ -14,3 +14,11 @@ export const closePopupsIfExist = async (page: Page) => {
     });
   }
 };
+
+export const getRowByName = async (page: Page, name: string) => {
+  if (await page.getByRole('button', { name: 'Clear filters' }).isVisible()) {
+    await page.getByRole('button', { name: 'Clear filters' }).click();
+  }
+  await page.getByPlaceholder(/^Filter by name.*$/).fill(name);
```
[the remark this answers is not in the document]
dominikvagner: sounds good, will improve it in that way 👍🏼😄

_playwright-tests/UI/IntrospectRepo.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+      await row.getByRole('gridcell', { name: '8', exact: true }).locator('a').click();
+      await expect(page.getByText('View list of packages')).toBeVisible();
+    });
+
+    await test.step('Check the modal for expected content', async () => {
+      await Promise.all([
+        expect(page.getByText('noarch')).toHaveCount(8),
+        expect(page.getByText('cheetah')).toBeVisible(),
+        expect(page.getByText('0.3').first()).toBeVisible(),
+        expect(page.getByText('0.8').first()).toBeVisible(),
```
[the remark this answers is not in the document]
dominikvagner: the repo is currently under katello fakerepos, but we are using those fakerepos in several tests, once when the changes by Ryan are done, then we could change all repos we use in tests everywhere, but for now I would leave it as is 😅 

the checks just verify that the 8 packages present in the repo are displayed and the one of them, cheetah, has the correct attributes (version, release, arch), I am not sure how to better test that the introspected repo has the correct packages in our UI 😄

_playwright-tests/UI/IntrospectRepo.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    await test.step('Wait for status to be "Valid"', async () => {
+      const row = await getRowByName(page, repoName);
+      await expect(row.getByText('Valid')).toBeVisible();
+    });
+  });
+
+  test('Check introspected repository packages', async ({ page }) => {
+    await test.step('Open the packages modal', async () => {
+      const row = await getRowByName(page, repoName);
+      await row.getByRole('gridcell', { name: '8', exact: true }).locator('a').click();
```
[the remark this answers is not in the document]
dominikvagner: I had some troubles choosing a good selector for this so I used the playwright inspector after I paused the test and this was the selector it has given me for the element.
but I think it's not a bad way of targeting this, it checks for a cell in the row that has 8 wrapped in a link inside, which should be there because in this way we show the count of packages in the repo and link to the modal 😅 
but maybe I should make constant like 'repoPakcageCount' so that the 8 is not just a magic number 😄

_playwright-tests/UI/IntrospectRepo.spec.ts
marusak: Huh, why are there two button with the same text? :D

_playwright-tests/UI/IntrospectRepo.spec.ts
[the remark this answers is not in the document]
marusak: Do you really check that cheetah has the correct attributes, or that any of those packages has those expected attributes? From the test it does not seem you enforce that cheetah has those values.

_playwright-tests/UI/IntrospectRepo.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+      await row.getByLabel('Kebab toggle').first().click();
+      await row.getByRole('menuitem', { name: 'Delete' }).click();
+      await expect(page.getByText('Remove repositories?')).toBeVisible();
+
+      await Promise.all([
+        page.waitForResponse(
+          (resp) =>
+            resp.url().includes('bulk_delete') && resp.status() >= 200 && resp.status() < 300,
+        ),
+        page.getByRole('button', { name: 'Remove' }).click(),
```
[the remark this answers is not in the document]
dominikvagner: added 🫡

_playwright-tests/UI/IntrospectRepo.spec.ts
dominikvagner: well there is the one at the top which is always visible and one specifically in the empty state below some help text, we have this way on both repos and templates pages (proposed by UX, at least it's in the mocks), so the `first()` has to be here 😄

_playwright-tests/UI/IntrospectRepo.spec.ts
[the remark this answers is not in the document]
dominikvagner: you're right, fixed 👍🏼

_playwright-tests/UI/helpers/helpers.ts
```diff
@@ -14,3 +14,45 @@ export const closePopupsIfExist = async (page: Page) => {
...
 };
+export const filterByName = async (page: Page, name: string) => {
+  await page.getByPlaceholder(/^Filter by name.*$/).fill(name);
+};
+
+export const clearFilters = async (page: Page) => {
+  try {
+    await page.getByRole('button', { name: 'Clear filters' }).waitFor({ timeout: 5000 });
+  } catch {
+    return Promise<void>;
```
Andrewgdewar, line 25: This seems like a strange return type. 
Would return; not work here?

_playwright-tests/UI/IntrospectRepo.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+      await row.getByRole('gridcell', { name: repoPackageCount, exact: true }).locator('a').click();
+      await expect(page.getByText('View list of packages')).toBeVisible();
+      await expect(page.getByText('Version').first()).toBeVisible();
+    });
+
+    await test.step('Check the modal for expected content', async () => {
+      const row = page.getByRole('row').filter({ has: page.getByText(testPackage) });
+      await Promise.all([
+        expect(page.getByText(repoArch)).toHaveCount(8),
+        expect((await getRowCellByHeader(row, 'Name')).getByText(testPackage)).toBeVisible(),
```
Andrewgdewar, line 66: is the double (()) needed here?

_playwright-tests/UI/IntrospectRepo.spec.ts
```diff
@@ -0,0 +1,94 @@
+import { test, expect } from '@playwright/test';
+import { navigateToRepositories } from './helpers/navHelpers';
+import { closePopupsIfExist, getRowByName, getRowCellByHeader } from './helpers/helpers';
+import { deleteAllRepos } from './helpers/deleteRepositories';
+
+test.describe('Introspect Repositories', () => {
+  const repoName = 'introspection-test';
+  const repoUrl = 'https://fedorapeople.org/groups/katello/fakerepos/zoo/';
```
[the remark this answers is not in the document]
rverdile: this has to merge before we can start using them: https://github.com/content-services/fixtures/pull/1

this could likely be a drop-in replacement in the meantime: https://rverdile.github.io/fixtures/yum/comps-modules/v1/, but not against leaving this for now

_playwright-tests/UI/IntrospectRepo.spec.ts
dominikvagner: yes it is

_playwright-tests/UI/helpers/helpers.ts
dominikvagner: it probably would, this just explicitly specifies the type instead of it being inferred, but I can change that if desired 😅

_playwright-tests/UI/IntrospectRepo.spec.ts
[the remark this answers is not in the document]
dominikvagner: I would leave this as is for now, and then change all repos at once after we have our own 😅

playwright.config.ts
```diff
@@ -20,8 +20,8 @@ export default defineConfig({
         ['./ci-reporter'],
       ]
     : 'list',
-  timeout: process.env.CI ? 60000 : 30000,
-  expect: { timeout: process.env.CI ? 30000 : 20000 },
+  timeout: 900_000,
+  expect: { timeout: process.env.CI ? 30_000 : 20_000 },
```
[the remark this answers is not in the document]
dominikvagner: the `_` is there just for better readability. 👓 

@marusak and I talked about this yesterday and yes it's a lot, but it's on purpose, the total test timeout shouldn't matter if there are the 'expect', 'action' and 'navigation' timeouts, it would be better to rely on the most granular timeouts and if needed just increase those for the singular units/steps 😅 
of course I am open to other ways of going about this, would be a good topic for the PW retro 😄 

adjusted the timeouts again and lowered this one a bit so have a look 👀

_playwright-tests/UI/helpers/helpers.ts
dominikvagner: changed 🫡

playwright.config.ts
```diff
@@ -20,9 +20,12 @@ export default defineConfig({
...
       ]
     : 'list',
-  timeout: process.env.CI ? 60000 : 30000,
-  expect: { timeout: process.env.CI ? 30000 : 20000 },
+  globalTimeout: 29.5 * 60 * 1000, // Set because of codebuild, we want PW to timeout before CB to get the results.
+  timeout: 10 * 60 * 1000,
+  expect: { timeout: process.env.CI ? 30_000 : 20_000 },
   use: {
+    actionTimeout: 30_000,
+    navigationTimeout: 30_000,
```
marusak, line 28: 5 different timeouts :D 

I would appreciate comment in the code on what is `timeout`, how does `expect` differ from `actionTimeout`.

And maybe let's just define all of them as 30 000? I think the `CI` check might not be necessary - slow computer or slow network at home might be worse that CI :D

dominikvagner: That global one is just because of codebuild and results, that's easy. And if want more granular timeouts, we have to set all of those that are more down the line from the test one (i.e. that's just the `timeout`), because defaults for those are 0 which means unlimited. 😄 
Yeah the CI might not be needed in the expect one 😅 .

Andrewgdewar: We are going to have a lot of people working in this repo, I'd suggest just changing all of these values to their int equivalents to simplify this. Maybe simply have the number followed by the comment of the time it equates to.

Simplicity is king. 

number // x minutes

PR #445, opened 2025-02-18
Test: Use currents.dev as reporter for Playwright

marusak:
Using gh-pages is simple but it has couple of issues:
- We need to have publishing workflow
- Maximum file size we can publish is 100M, but some tests already exceed it
- There is limit on how many logs we can have at the same time
- We need to have cleanup action
- We don;t have any historical data

We can use currents.dev It has multiple advantages:
- It is service, we literally only need to set up two enviroment variables and we are good to go
- Unlimited space, we can even have videos
- There is full history of all runs. You can see metrics on test runs, duration, flakiness, which tests fail the most all in one dashboard

It has two disadvantages:
- Test results are not as clear as I would like it (you need to drill deeper to find out logs)
- You need to login to see those results

It can also post comments and status checks, see https://github.com/marusak/pw-test/pull/8 (also check 4 successful checks)

I can demo how the dashboard looks like. I also tested this and it worked, you can see the run here: https://github.com/content-services/content-sources-frontend/actions/runs/13391757077/job/37400853196 and test results here: https://app.currents.dev/run/4a446233cead3582 (First I need to give you access)


.github/workflows/playwright-actions.yml
```diff
@@ -206,65 +206,16 @@ jobs:
...
           report-path: './playwright-ctrf/playwright-ctrf.json'
         if: ${{ !cancelled() }}
-
-      - name: Store front-end Test report
-        uses: actions/upload-artifact@v4
-        if: ${{ !cancelled() }}
-        with:
-          name: playwright-report
-          path: playwright-report/
-          retention-days: 10
```
marusak, line 227: Maybe I should keep this for publicly accessible report

playwright.config.ts
```diff
@@ -12,12 +12,11 @@ export default defineConfig({
   workers: 1,
   reporter: process.env.CI
     ? [
-        ['html', { outputFolder: 'playwright-report' }],
```
marusak, line 15: and this

.github/workflows/playwright-actions.yml
marusak: done

Andrewgdewar (approved):
Merging, will rollback if not functioning.

PR #446, opened 2025-02-18
Test: Allow triggering PW tests from main or manually

marusak:
Splitting out from #445


marusak:
I split this because I don't think we should squash these changes with currents PR. If we enable rebasing, then we can close this one and just look at #445

PR #449, opened 2025-02-20
Build: Use labels correctly in FormGroup

marusak:
We were not able to target some elements by label. This showed to be actually bug in our code which made some elements inaccessible. According to documentation [1]:
  fieldId: ID of an individual field or a group of multiple fields.
           Required when a role of "group" or "radiogroup" is passed in.
           If _only one field_ is included, its ID attribute and this prop _must be the same_.

[1] https://www.patternfly.org/components/forms/form/#formgroup


Andrewgdewar:
Introspect test failures expected until the merge of Ryan's PR.

Andrewgdewar (approved):

PR #452, opened 2025-02-24
Test: improve PW tests flakiness

dominikvagner:
## Summary
This PR changes/fixes a few things to combat test flakiness.
- Changes a step in the PW actions from processing all repos to only snapshot the RH one that is used for tests. (as in https://github.com/content-services/content-sources-backend/pull/985)
- Increases the timeout for the action that waits on the introspection.
- Introduces 2 new helpers, one for retrying navigation (which sometimes just fails/times-out, because of console-dot) and another one which waits for a specific task to be picked up.
- Also a custom ESLint rule was added that will throw linting errors when any local import is made inside of a PW test and imports anything outside the PW test folder. (This will be helpful for preventing failing of our integration PW tests)


_playwright-tests/UI/CustomRepo.spec.ts
```diff
@@ -1,15 +1,15 @@
 import { test, expect, type Page } from '@playwright/test';
 import { navigateToRepositories } from './helpers/navHelpers';
 import { deleteAllRepos } from './helpers/deleteRepositories';
-import { closePopupsIfExist } from './helpers/helpers';
+import { closePopupsIfExist, retry } from './helpers/helpers';
```
marusak, line 4: Although controversial, I agree with this :) I know that consoleDot sometimes gets stuck and I need to reload for it to load. But should we just not move this into the navigation function rather than introducing this to every single test? Would be cleaner and we would not forget to introduce it in other tests

_playwright-tests/UI/helpers/helpers.ts
```diff
@@ -57,6 +57,33 @@ export const getRowCellByHeader = async (row: Locator, name: string) => {
...
+export const waitForTaskPickup = async ({ request }: Page, repoUrl: string, type: string) => {
+  const response = await request.get(`/api/content-sources/v1/repositories/?url=${repoUrl}`);
+  expect(response.status()).toBe(200);
+  const body = await response.json();
+  expect(Array.isArray(body.data)).toBeTruthy();
+  const uuidList = body.data.map((data: { uuid: string }) => data.uuid) as string[];
+  expect(uuidList.length).toEqual(1);
+
+  const repoUuid = uuidList[0];
+  while (true) {
```
marusak, line 75: I am bit torn on this one. It might be necessary for stage/prod tests as we are not in control of queue or for parallel tests, but I feel that this might hide actual issues or make tests unpredictable. The golden rule of tests - if they pass, none cares about them would apply here, as if they pass but always wait for very long time, we would not notice it.

_playwright-tests/UI/RedHatRepo.spec.ts
```diff
@@ -28,7 +27,7 @@ test.describe('Red Hat Repositories', () => {
         });
 
         await test.step('Wait for a valid status', async () => {
-            await status.waitFor({ state: 'visible' });
+            await status.waitFor({ state: 'visible', timeout: 600_000 });
```
marusak: Very good! Timeouts should be as specific and minimal as possible and it should be clear why we need extra timeout.

marusak:
Nice, thanks! I have some comments, but in general I agree and like those changes. And nice work on commits ;) Was very easy to review commit by commit!
Have you tried running action on this to see if it fixes that one stubborn test?

_playwright-tests/UI/helpers/helpers.ts
dominikvagner: I am also a bit torn on this as it's sort of a foot gun for the reason you mentioned, but on the other hand the queue being bogged down on the runner isn't supposed to be the thing this test is verifying so I included this for that reason. 😅 
but I am open to this being axed from this test (or the entire helper) @Andrewgdewar what ya think? 👀

_playwright-tests/UI/helpers/helpers.ts
```diff
@@ -10,7 +10,11 @@ export const closePopupsIfExist = async (page: Page) => {
 
   for (const locator of locatorsToCheck) {
     await page.addLocatorHandler(locator, async () => {
-      await locator.first().click(); // There can be multiple toast pop-ups
+      try {
+        await locator.first().click(); // There can be multiple toast pop-ups
```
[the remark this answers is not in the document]
dominikvagner: hrm, you are right, I put the wrong reason in the commit description mixed stuff up 😵‍💫 
this actually fixes a different problem, which is that multiple handlers try to close the same popup, but one second one fails as it's already closed, will change the description 😅

_playwright-tests/UI/CustomRepo.spec.ts
dominikvagner: forgetting this can be an issue, but I didn't want to force it into existing helpers as I think it's mostly wanted only on the inital page load (and had some trouble trying this while prototyping) so I just added a new one, but it could be changed 😅 
@Andrewgdewar what's your opinion, retry included in nav helpers or it needs to be explicitly used when desired?

dominikvagner:
> Nice, thanks! I have some comments, but in general I agree and like those changes. And nice work on commits ;) Was very easy to review commit by commit! Have you tried running action on this to see if it fixes that one stubborn test?

thanks 🎉😇 
that one won't pass until the `snapshot` command will also introspect, [PR is open](https://github.com/content-services/content-sources-backend/pull/995) for that 😅

_playwright-tests/UI/CustomRepo.spec.ts
Andrewgdewar: I think moving it to the navigation functions makes sense, and keeping the helper around in case we create other navigation based things!

_playwright-tests/UI/helpers/helpers.ts
```diff
@@ -56,3 +60,50 @@ export const getRowCellByHeader = async (row: Locator, name: string) => {
 
   return row.getByRole('gridcell').nth(index);
 };
+
+export const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
+
+export const waitForTaskPickup = async ({ request }: Page, repoUrl: string, type: string) => {
```
Andrewgdewar, line 66: I'd say make the timeout a passed in variable here.

_playwright-tests/UI/helpers/helpers.ts
```diff
@@ -56,3 +60,50 @@ export const getRowCellByHeader = async (row: Locator, name: string) => {
...
+  const response = await request.get(`/api/content-sources/v1/repositories/?url=${repoUrl}`);
+  expect(response.status()).toBe(200);
+  const body = await response.json();
+  expect(Array.isArray(body.data)).toBeTruthy();
+  const uuidList = body.data.map((data: { uuid: string }) => data.uuid) as string[];
+  expect(uuidList.length).toEqual(1);
+
+  const repoUuid = uuidList[0];
+  while (true) {
```
Andrewgdewar, line 75: If I am reading this correctly we could wait forever.... also the timeout on the request below is excessive, I'd expect that to return in 5s or so? 
Why would an api for tasks call take 60 seconds...

I like the idea of what you are doing here but lets have this while loop be based off off of a retry count, maybe configurable via passed in variable with a reasonable number set by default. maybe check this 10 times at maximum for example.

_playwright-tests/UI/helpers/helpers.ts
```diff
@@ -56,3 +60,50 @@ export const getRowCellByHeader = async (row: Locator, name: string) => {
...
+    );
+    const body = await response.json();
+    const data = Array.from(body.data);
+    if (data.length === 0) {
+      break;
+    }
+    await sleep(5000);
+  }
+
+  return;
```
Andrewgdewar, line 88: unneeded return

_playwright-tests/UI/helpers/helpers.ts
```diff
@@ -56,3 +60,50 @@ export const getRowCellByHeader = async (row: Locator, name: string) => {
...
+      try {
+        await callback(page);
+      } catch {
+        continue;
+      }
+      break;
+    }
+  }
+
+  return;
```
Andrewgdewar, line 108: this is the same as not having a return, I guess some prefer explicit returns. As long as you are away 🤷

_playwright-tests/UI/CustomRepo.spec.ts
dominikvagner: changed!

_playwright-tests/UI/helpers/deletePopularRepositories.ts
```diff
@@ -29,7 +29,5 @@ export const deleteAllPopularRepos = async ({ request }: Page) => {
       console.error('Failed to delete repositories:', error);
       throw error;
     }
-  } else {
-    console.log('No repositories to delete.');
```
marusak, line 33: Why is this deleted?

_playwright-tests/UI/RedHatRepo.spec.ts
```diff
@@ -52,4 +52,4 @@ test.describe('Red Hat Repositories', () => {
             await page.getByTestId('snapshot_list_modal-ModalBoxCloseButton').click();
         });
     });
-});
```
marusak, line 55: let's try avoid commiting these :) (I agree there should be trailing new line, but if someone commits it, usually just changing it is meh - but we can introduce lint for it)

_playwright-tests/UI/helpers/helpers.ts
```diff
@@ -57,6 +57,31 @@ export const getRowCellByHeader = async (row: Locator, name: string) => {
...
+export const waitForTaskPickup = async (page: Page, repoUrl: string, type: string) => {
+  const response = await page.request.get(`/api/content-sources/v1/repositories/?url=${repoUrl}`);
+  expect(response.status()).toBe(200);
+  const body = await response.json();
+  expect(Array.isArray(body.data)).toBeTruthy();
+  const uuidList = body.data.map((data: { uuid: string }) => data.uuid) as string[];
+  expect(uuidList.length).toEqual(1);
+  const repoUuid = uuidList[0];
+
+  await expect.poll(async () => {
```
marusak, line 75: cool approach!

marusak (approved):
Just two nitpicks in the first commit :)

marusak (changes requested):
lint is failing, please take a look https://github.com/content-services/content-sources-frontend/actions/runs/13765169503/job/38489828892?pr=452

_playwright-tests/UI/helpers/deletePopularRepositories.ts
dominikvagner: because it unnecessarily spams the PW test report and given that this function its used for auto cleanup this log isn't useful 😅

_playwright-tests/UI/RedHatRepo.spec.ts
dominikvagner: fixed 👍🏼

_playwright-tests/UI/helpers/helpers.ts
```diff
@@ -79,3 +83,59 @@ export const validateSnapshotTimestamp = async (timestamp: string, howRecent: nu
...
+        timeout: 300_000, // 5 min
+      },
+    )
+    .toBeTruthy();
+};
+
+export const retry = async (
+  page: Page,
+  callback: (page: Page) => Promise<void>,
+  tries = 3,
```
Andrewgdewar, line 120: 👍

Andrewgdewar (approved):

PR #454, opened 2025-03-03
Test: process single repo instead of all in pw ci

dominikvagner:
## Summary
This changes GH action `playwright-actions` to snapshot only a single repo instead of running process-repos command for all repos. This is needed as the process-repos bogs down the test runner and tests fail/timeout because of that.


marusak (approved):
Thanks!
Please run it manually (as it changes gh action file), and if it works feel free to merge, thanks!

PR #457, opened 2025-03-04
Test: fix get row cell by header helper

dominikvagner:
## Summary
This adds a `toBeVisible` expect to the `getRowCellByHeader` helper which should help with preventing the current flakiness of this helper, caused by slow loading and the presence of the skeleton/zero-state table.


marusak (approved):
Thanks!

PR #459, opened 2025-03-05
HMS-5441: replace custom repositories test with snapshot test

tkoscieln:
## Summary
Modify the CustomRepositories test to verify repo snapshotting. Includes refactoring of the code by introducing test steps and simplifying the test structure to a single test without additional functions.

## Testing steps
* tests should pass
* test is supposed to behave as described in the Jira ticket


jlsherrill:
https://issues.redhat.com/browse/HMS-5441

_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+    await test.step('Filter by version', async () => {
+      const versionFilterButton = page.getByRole('button', { name: 'filter version' });
+      await versionFilterButton.click();
+      await page.getByRole('menuitem', { name: 'el9' }).locator('label').click();
+      await page.getByRole('menuitem', { name: 'el8' }).locator('label').click();
+      await versionFilterButton.click(); // Close the version filter dropdown
+    });
+
+    await test.step('Wait for the successful API call', async () => {
+      if (await page.locator('.pf-v5-c-helper-text__item.pf-m-error').isVisible()) {
```
marusak, line 50: This is racy by design. This might pass just fine, because when this is being checked there might not be any error yet. If you want check like this, you want to add listener, similar to https://github.com/content-services/content-sources-frontend/blob/main/_playwright-tests/UI/helpers/helpers.ts#L12

_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+        expect(page.locator('div[id^="pf-modal-part"]').first()).not.toBeVisible(),
+      ]);
+    });
+
+    await test.step('Verify that snapshot is successful', async () => {
+      await expect(page.getByText('Valid')).toBeVisible({ timeout: 60000 });
+    });
+
+    await test.step('Verify that snapshot is in snapshots list', async () => {
+      await page.getByRole('button', { name: 'Kebab toggle' }).first().click();
```
marusak, line 72: What a useless name :) (not related, just find it funny as someone who cannot see would not find this helpful :D )

_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+      await page.getByRole('menuitem', { name: 'View all snapshots' }).click();
+      await expect(page.getByLabel('SnapshotsView list of').locator('tbody')).toBeVisible();
```
marusak, line 74: Interesting label :man_shrugging:

_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+        .locator('tbody')
+        .textContent();
+      if (snapshotTimestamp != null) {
+        if ((await validateSnapshotTimestamp(snapshotTimestamp, 10)) == false) {
+          throw new Error('Most recent snapshot timestamp is older than 10 minutes!');
+        }
+      } else {
+        throw new Error('Snapshot timestamp not found!');
+      }
+      await page.getByLabel('Close', { exact: true }).click();
```
marusak, line 86: Should this wait until the dialog closes? E.g. some locator you used before is not visible?

tkoscieln: I think so, because the delete step that follows needs to select the row in repositories list and it needs to open the delete modal.

_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+    });
+
+    await test.step('Delete created repository', async () => {
+      const row = await getRowByName(page, repoName);
```
dominikvagner, line 90: here it would be better to use an url for the repo filtering 🔧 
the `getRowByName()` can also accept URL and it is better to filter by that rather than the name (unless the name is an UUID) 😅 
- _because: if you have 2 repos with names 'one' and 'test one', it will return list both and thus the kebab click will throw, because it won't know which one to select_

marusak: >  'one' and 'test one', it will return list both

Maybe we want to use `exact: true`? :thinking:

PR #463, opened 2025-03-07
HMS-5618: Explore parallel tests, fix issues around idempotency.

Andrewgdewar:
## Summary

-  Makes tests run in parallel locally only
- Fixes a number of tests that were affecting other tests (namely their cleanup functions

Additional changes: Fixed a number of issues with helpers, namely around not waiting for filters to take effect before continuing.

### DIFF
--- a/_playwright-tests/UI/IntrospectRepo.spec.ts
+++ b/_playwright-tests/UI/IntrospectRepo.spec.ts
@@ -95,10 +95,8 @@ test.describe('Introspect Repositories', () => {
         ),
         page.getByRole('button', { name: 'Remove' }).click(),
       ]);
-
-      await expect(
-        page.getByText('No custom repositories match the filter criteria'),
-      ).toBeVisible();
+      // Ensure the specific row is removed
+      await expect(row).not.toBeVisible();
     });
   });
 });
--- a/_playwright-tests/UI/IntrospectRepo.spec.ts
+++ b/_playwright-tests/UI/IntrospectRepo.spec.ts
@@ -66,9 +66,10 @@ test.describe('Introspect Repositories', () => {
     });
 
     await test.step('Check the modal for expected content', async () => {
-      const row = page.getByRole('row').filter({ has: page.getByText(testPackage) });
+      const modal = page.getByTestId('rpm_package_modal');
+      const row = modal.getByRole('row').filter({ has: page.getByText(testPackage) });
       await Promise.all([
-        expect(page.getByText(repoArch)).toHaveCount(8),
+        expect(modal.locator('tbody')).toHaveCount(8),
         expect((await getRowCellByHeader(page, row, 'Name')).getByText(testPackage)).toBeVisible(),
         expect(
           (await getRowCellByHeader(page, row, 'Version')).getByText(repoVersion),
--- a/_playwright-tests/UI/PopularRepo.spec.ts
+++ b/_playwright-tests/UI/PopularRepo.spec.ts
@@ -1,6 +1,6 @@
 import { test, expect } from '@playwright/test';
 import { navigateToRepositories } from './helpers/navHelpers';
-import { closePopupsIfExist } from './helpers/helpers';
+import { closePopupsIfExist, getRowByNameOrUrl } from './helpers/helpers';
 import { deleteAllPopularRepos } from './helpers/deletePopularRepositories';
 
 test.describe('Popular Repositories', () => {
@@ -20,7 +20,7 @@ test.describe('Popular Repositories', () => {
 
     await test.step('Select EPEL 9', async () => {
       await page
-        .getByRole('row', { name: 'EPEL 9 Everything x86_64' })
+        .getByRole('row', { name: repoName })
         .getByLabel('Select row 0', { exact: true })
         .click();
       await page.getByTestId('add-selected-dropdown-toggle-no-snap');
@@ -45,7 +45,7 @@ test.describe('Popular Repositories', () => {
       ).toBeVisible();
       await expect(
         page
-          .getByRole('row', { name: 'EPEL 9 Everything x86_64' })
+          .getByRole('row', { name: repoName })
           .getByTestId('remove_popular_repo')
           .getByText('Remove'),
       ).toBeVisible();
@@ -77,7 +77,8 @@ test.describe('Popular Repositories', () => {
 
     await test.step('Use kebab menu to delete a repo', async () => {
       await page.getByRole('textbox', { name: 'Filter by name/url' }).fill('EPEL');
-      await page.getByRole('checkbox', { name: 'Select row 0' }).check();
+      const row = await getRowByNameOrUrl(page, repoName);
+      await row.getByRole('checkbox', { name: 'Select row 0' }).check();
       await page.getByTestId('custom_repositories_kebab_toggle').click();
       await page.getByRole('menuitem', { name: 'Remove 1 repositories' }).click();
       // Confirm the removal in the pop-up
--- a/_playwright-tests/UI/PopularRepo.spec.ts
+++ b/_playwright-tests/UI/PopularRepo.spec.ts
@@ -3,6 +3,9 @@ import { navigateToRepositories } from './helpers/navHelpers';
 import { closePopupsIfExist, getRowByNameOrUrl } from './helpers/helpers';
 import { deleteAllPopularRepos } from './helpers/deletePopularRepositories';
 
+const repoName9 = 'EPEL 9 Everything x86_64';
+const repoName8 = 'EPEL 8 Everything x86_64';
+
 test.describe('Popular Repositories', () => {
   test('Add popular repos', async ({ page }) => {
     // Ensure no popular repos are selected.
@@ -20,7 +23,7 @@ test.describe('Popular Repositories', () => {
 
     await test.step('Select EPEL 9', async () => {
       await page
-        .getByRole('row', { name: repoName })
+        .getByRole('row', { name: repoName9 })
         .getByLabel('Select row 0', { exact: true })
         .click();
       await page.getByTestId('add-selected-dropdown-toggle-no-snap');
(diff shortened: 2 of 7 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-5618

_playwright-tests/UI/IntrospectRepo.spec.ts
```diff
@@ -64,8 +64,12 @@ test.describe('Introspect Repositories', () => {
...
         expect(page.getByText(repoArch)).toHaveCount(8),
         expect((await getRowCellByHeader(page, row, 'Name')).getByText(testPackage)).toBeVisible(),
-        expect((await getRowCellByHeader(page, row, 'Version')).getByText(repoVersion)).toBeVisible(),
-        expect((await getRowCellByHeader(page, row, 'Release')).getByText(repoRelease)).toBeVisible(),
+        expect(
+          (await getRowCellByHeader(page, row, 'Version')).getByText(repoVersion),
+        ).toBeVisible(),
+        expect(
+          (await getRowCellByHeader(page, row, 'Release')).getByText(repoRelease),
+        ).toBeVisible(),
```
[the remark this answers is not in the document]
Andrewgdewar: There are a lot of lint changes, when contributors don't use standardized tooling, this will exist.

yarn.lock
```diff
@@ -53,7 +53,7 @@
     "@jridgewell/gen-mapping" "^0.3.5"
     "@jridgewell/trace-mapping" "^0.3.24"
 
-"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.10.4", "@babel/code-frame@^7.12.13", "@babel/code-frame@^7.16.7", "@babel/code-frame@^7.25.9", "@babel/code-frame@^7.26.2":
+"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.10.4", "@babel/code-frame@^7.12.13", "@babel/code-frame@^7.16.7", "@babel/code-frame@^7.18.6", "@babel/code-frame@^7.25.9", "@babel/code-frame@^7.26.2":
```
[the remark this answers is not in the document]
marusak: Bump? (why was this comment resolved?)

_playwright-tests/UI/helpers/helpers.ts
```diff
@@ -20,6 +20,7 @@ export const closePopupsIfExist = async (page: Page) => {
 };
 export const filterByNameOrUrl = async (page: Page, name: string) => {
   await page.getByPlaceholder(/^Filter by name.*$/).fill(name);
+  expect(await page.getByText(name).count()).toBeGreaterThan(1);
```
marusak, line 22: Do we actually want this to be more than one? :thinking: Or do we want the filter to have just one result?

Andrewgdewar: When we filter something, there are 3 components that can appear on the screen.
The chip, the row in the table, and the text in the input. 
The test picks up the chip only half the time, if the filter text is short enough. 

I suppose a better solution would be to use a locator reduction on the table, and check within the rows themselves...

... Just because you are right doesn't mean I like it... 😝

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -80,11 +80,8 @@ test.describe('Popular Repositories', () => {
     });
 
     await test.step('Use kebab menu to delete a repo', async () => {
-      const filterInput = page.getByRole('textbox', { name: 'Filter by name/url' });
-      // We need to remember to clear the filter after specifying it above!
-      await filterInput.clear();
-      // Now we can enter a new value...
-      await filterInput.fill(repoName);
+      const searchInput = page.getByRole('textbox', { name: 'Filter by name/url' });
```
marusak, line 83: ouf of interestes, how is this helping with `Fix test hang on popular repo delete ` :thinking: (adding commit message would be useful).

You technically removed the clearing of the input - was there issue with it?

_playwright-tests/UI/helpers/navHelpers.ts
```diff
@@ -34,12 +34,10 @@ export const navigateToRepositories = async (page: Page) => {
...
   await page.goto('/insights/content/templates');
 
-  const repositoriesListPage = page.getByText(
-    'View all content templates within your organization.',
-  );
+  const templateText = page.getByText('View all content templates within your organization.');
 
   // Wait for either list page or zerostate
-  await repositoriesListPage.waitFor({ state: 'visible' });
+  await templateText.waitFor({ state: 'visible' });
```
marusak, line 38: Hmm, is this change just to keep the variable shorter so it does not wrap the command? :D

marusak (approved):
Thanks!
Some of those changes are bit questionable (e.g. renaming variable for no obvious reason without providing explanation in commit messages) or one commit introducing variable, using it in some places and then different commit using that same variable in the rest of the code.
One excuse is that a lot of this is still in flux and we need to standardize on what is acceptable so let's get this in without too much nitpicking :)

_playwright-tests/UI/PopularRepo.spec.ts
Andrewgdewar: The clear took 20ms, so just re-filling was faster, which prevented an occasional test failure.

_playwright-tests/UI/helpers/navHelpers.ts
Andrewgdewar: I think the name change was meant to better describe the selector.

Andrewgdewar: But I suppose your point was the commit wasn't clear.

_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -64,7 +64,16 @@ test.describe('Snapshot Repositories', () => {
     });
 
     await test.step('Verify that snapshot is in snapshots list', async () => {
-      await page.getByRole('button', { name: 'Kebab toggle' }).click();
+      const row = await getRowByNameOrUrl(page, repoName);
+      await row.getByLabel('Kebab toggle').click();
+      await page.waitForTimeout(500);
+      const menu = row.locator('.pf-v5-c-menu__content');
+      while (!(await menu.isVisible())) {
+        // Occasionally a re-render will case the menu to close unexpectedly
```
swadeley, line 72:
```suggestion
        // Occasionally a re-render will cause the menu to close unexpectedly
```

_playwright-tests/UI/helpers/navHelpers.ts
marusak: The commit only says "Parameter update". No commit message, so it is not clear why you did this change

swadeley: Hi, I agree that list of templates should not be called `repositoriesListPage`.
While `templateText` seems something specific to one template ( `templateListPage`?), I could not think of something better that would not be much longer.

dominikvagner (changes requested):
two small comments 👀😄
and I noticed some tests were failing locally when run in parallel instead of when run sequentially which is not ideal, so maybe we should make that the default yet? 😕😅 

otherwise good job fixing several of those small hard to find/fix things 👏🏼😇

_playwright-tests/UI/RedHatRepo.spec.ts
```diff
@@ -16,7 +16,8 @@ test.describe('Red Hat Repositories', () => {
     });
   });
 
-  test('Wait for repository status to be "Valid"', async ({ page }) => {
+  test('Wait for repository status to be "Valid"', async ({ page }, testinfo) => {
+    testinfo.setTimeout(600000);
```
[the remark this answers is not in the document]
Andrewgdewar: Nice catch!

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -52,8 +54,9 @@ test.describe('Popular Repositories', () => {
     });
 
     await test.step('Apply filter and clear it', async () => {
-      const searchInput = page.getByRole('textbox', { name: 'Filter by name/url' });
-      await searchInput.fill('EPEL 8 Everything x86_64');
+      await page
+        .getByRole('textbox', { name: 'Filter by name/url' })
+        .fill('EPEL 8 Everything x86_64');
```
dominikvagner, line 59: the string/name `'EPEL 8 Everything x86_64'` is used in several places in this file, why not use a variable instead? like for the EPEL 9 repo 🤔

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -76,7 +81,6 @@ test.describe('Popular Repositories', () => {
     });
 
     await test.step('Use kebab menu to delete a repo', async () => {
```
dominikvagner, line 83: why is this step named exactly the same as the one before it, the names should be different so it's clear which one failed 🧹 
and this whole step should ideally be parameterized in a for loop over the 2 repos that are being deleted 💭 ♻️
something like: 
```ts
const epelRepos = [epel9name, epel8name];
for (const repo of epelRepos) {
  await test.step(`Use kebab menu to delete the ${repo} repo`, async () => {
    const row = await getRowByNameOrUrl(page, repo);
    await page.getByRole('checkbox', { name: 'Select row 0' }).check();
    await page.getByTestId('custom_repositories_kebab_toggle').click();
    await page.getByRole('menuitem', { name: 'Remove 1 repositories' }).click();
    // Confirm the removal in the pop-up
    await page.getByRole('button', { name: 'Remove' }).click();
  });
}
```

dominikvagner (commented):
two small comments with some suggestions for improvements to the popular repos test ✨😅

_playwright-tests/UI/PopularRepo.spec.ts
marusak: Good idea but I think unrelated to this PR. Let's do that in a different PR

_playwright-tests/UI/PopularRepo.spec.ts
marusak: Fixed

dominikvagner (approved):
approved! good job @Andrewgdewar and @marusak 👏🏼🎉

marusak:
Green without retry!! (two flaky tests though, but very nice nevertheless!)

PR #464, opened 2025-03-11
Build: add PR head refs to github checkout actions

dominikvagner:
## Summary
This adds PR head refs to all GH checkout actions that currently don't have those, this is needed so that the actions run on the same code you are pushing and they don't have different results.


marusak (approved):
Thanks!

PR #481, opened 2025-03-18
HMS-5648: Add GPG key test

swadeley:
## Summary

Adding GPG keys to a repo 

## Testing steps

Tests pass.

Based on IQE test in associated Jira.


jlsherrill:
https://issues.redhat.com/browse/HMS-5648

_playwright-tests/UI/GPG_keys.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+      await page.getByRole('button', { name: 'Save', exact: true }).click();
+    });
+    await test.step('Wait for status to be "Valid"', async () => {
+      const row = await getRowByNameOrUrl(page, repoName);
+      await expect(row.getByText('Valid')).toBeVisible({ timeout: 60000 });
+    });
+    await test.step('Add a GPG Key', async () => {
+      // Search for the created repo
+      await page.getByRole('textbox', { name: 'Filter by name/url' }).fill(repoName);
+      const row = await getRowByNameOrUrl(page, repoName);
```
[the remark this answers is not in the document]
swadeley: ok, thanks

_playwright-tests/UI/GPG_keys.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+      await expect(row.getByText('Valid')).toBeVisible({ timeout: 60000 });
+      // Open edit modal
+      await row.getByLabel('Kebab toggle').click();
+      // Click on the Edit button to see the repo
+      await row.getByRole('menuitem', { name: 'Edit' }).click();
+      await expect(page.getByRole('dialog', { name: 'Edit custom repository' })).toBeVisible();
```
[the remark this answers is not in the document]
swadeley: No special reason. I can change it.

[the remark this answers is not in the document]
swadeley: changed

_playwright-tests/UI/GPG_keys.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+      await page.getByRole('button', { name: 'Save', exact: true }).click();
+    });
+    await test.step('Wait for status to be "Valid"', async () => {
+      const row = await getRowByNameOrUrl(page, repoName);
+      await expect(row.getByText('Valid')).toBeVisible({ timeout: 60000 });
+    });
+    await test.step('Add a GPG Key', async () => {
+      // Search for the created repo
+      const row = await getRowByNameOrUrl(page, repoName);
+      await expect(row.getByText('Valid')).toBeVisible({ timeout: 60000 });
```
marusak, line 43: These two commands are identical - lines 37-38 are the same as 42-43. Do we need it twice? :thinking:

_playwright-tests/UI/GPG_keys.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+      // Paste the URL to the packages GPG key
+      await page.getByPlaceholder('Paste GPG key or URL here').fill(packages_key);
+      // Save button would be disabled for bad or incorrect gpg key
+      await page.getByRole('button', { name: 'Save', exact: true }).click();
+    });
+    await test.step('Wait for status to be "Valid"', async () => {
+      const row = await getRowByNameOrUrl(page, repoName);
+      await expect(row.getByText('Valid')).toBeVisible({ timeout: 60000 });
+    });
+    await test.step('Add a GPG Key', async () => {
```
marusak, line 40: This is not "Add"-ing but "Edit"-ing.

_playwright-tests/UI/GPG_keys.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+      // Click on the 'Add repositories' button
+      // HMS-5268 There are two buttons on the ZeroState page
+      await page.getByRole('button', { name: 'Add repositories' }).first().click();
+      await expect(page.getByRole('dialog', { name: 'Add custom repositories' })).toBeVisible();
+      // Fill in the repository details
+      await page.getByLabel('Name').fill(`${repoName}`);
+      await page.getByLabel('Introspect only').click();
+      await page.getByLabel('URL').fill(url);
+      // Paste the URL to the packages GPG key
+      await page.getByPlaceholder('Paste GPG key or URL here').fill(packages_key);
```
marusak, line 32: This test should somehow verify that it has the correct key. Can you verify it somewhere that it uses the correct key? Do we see it somewhere? Or can we verify that we cannot access something without it and once we add it we can? Or something similar?
I wonder what is the value of this test - is it that we can enter value into this field or that if you enter the value you can do something.

_playwright-tests/UI/GPG_keys.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+      await expect(row.getByText('Valid')).toBeVisible({ timeout: 60000 });
+      // Open edit modal
+      await row.getByLabel('Kebab toggle').click();
+      // Click on the Edit button to see the repo
+      await row.getByRole('menuitem', { name: 'Edit' }).click();
+      await expect(page.getByRole('dialog', { name: 'Edit custom repository' })).toBeVisible();
+      // Paste the URL to the repo metadata GPG key
+      await page.getByPlaceholder('Paste GPG key or URL here').fill(meta_key);
+      // Save button would be disabled for bad or incorrect gpg key
+      await page.getByRole('button', { name: 'Save changes', exact: true }).click();
```
marusak, line 52: What is value of this test case? Is it that you can click on save button after you change gpg key? Because it does not validate that for example the change was successful - nor in UI, nor that it actually changes something for those repositories.

marusak (changes requested):
Very nicely written test, but I have question of the use - should these tests also verify that changes that we do actually have some real effect?

swadeley:
> Very nicely written test, but I have question of the use - should these tests also verify that changes that we do actually have some real effect?

The purpose of the test is to verefy the keys can be stored for use by clients. The console content-sources app makes some basic checks on the format of the keys when you add them. When a client adds a repo it can get the key and use it verify packages or repo metadata.

_playwright-tests/UI/GPG_keys.spec.ts
swadeley: If the key was not in a valid format the Save button would not be enabled. 
Our test just shows we can enter a key and it is checked for valid format and saved.

It is up to the user or clients to use, or not, the key as they see fit.

_playwright-tests/UI/GPG_keys.spec.ts
swadeley: Partly to prove we can change the key, partly to check the buttons and validation logic for the different key.
I can add a step to read it back to ensure it was changed.

_playwright-tests/UI/GPG_keys.spec.ts
swadeley: Changed

_playwright-tests/UI/GPG_keys.spec.ts
swadeley: Removed 37 to 38

_playwright-tests/UI/GPG_keys.spec.ts
swadeley: I have added a line to assert there is a GPG token in the field after pasting the URL.

_playwright-tests/UI/GPG_keys.spec.ts
swadeley: I have add a check that validate fails if you select the metadata option for a package verification only key.

_playwright-tests/UI/GPG_keys.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+import { closePopupsIfExist, getRowByNameOrUrl } from './helpers/helpers';
+
+export const repoNamePrefix = 'GPG-key';
+export const randomName = () => (Math.random() + 1).toString(36).substring(2, 6);
+export const repoName = `${repoNamePrefix}-${randomName()}`;
+
+export const url = 'https://jlsherrill.fedorapeople.org/fake-repos/signed/';
+export const packages_key =
+  'http://jlsherrill.fedorapeople.org/fake-repos/needed-errata/RPM-GPG-KEY-dummy-packages-generator';
+export const meta_key = 'https://jlsherrill.fedorapeople.org/fake-repos/signed/GPG-KEY.gpg';
```
dominikvagner, line 13: none of these consts should be exported, unless you are using them in other test files and are generic enough 💭 (in that case, they should be moved to helpers)
this seems like test data for only this test, so no need to export/make them public 😅

_playwright-tests/UI/GPG_keys.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+      // Click on the 'Add repositories' button
+      // HMS-5268 There are two buttons on the ZeroState page
+      await page.getByRole('button', { name: 'Add repositories' }).first().click();
+      await expect(page.getByRole('dialog', { name: 'Add custom repositories' })).toBeVisible();
+      // Fill in the repository details
+      await page.getByLabel('Name').fill(`${repoName}`);
+      await page.getByLabel('Introspect only').click();
+      await page.getByLabel('URL').fill(url);
+      // Paste the URL to the packages GPG key
+      await page.getByPlaceholder('Paste GPG key or URL here').fill(packages_key);
```
dominikvagner, line 32: _small style nit:_ does the comment above this line need to be here? given that we use the user first test actions/locators the actual line it "documents" reads exactly the same as the comment above it, so it doesn't add much value I think, feel free to disagree with me on this 😅 

_(this could be said about more comments in this file)_
_(but if you'd like/prefer to have those comments there feel free to ignore this 😄)_

dominikvagner (commented):
one small comment about the exports in this file and one tiny style nit 👀😅

otherwise this test looks very nice ✨ 
good job! 👏🏼😇

_playwright-tests/UI/GPG_keys.spec.ts
swadeley: OK, those likes are test specific , will remove the `export`

_playwright-tests/UI/GPG_keys.spec.ts
swadeley: Ok, old habits die hard (using in-line comments). I'll remove it.

dominikvagner: haha, yeah that's true 😄 there is still place for some, before more actions grouped together when it can be hard to easily figure out what's happening, or ones that link to tickets (like you did in this PR) 😌

thanks for looking into this 🙇🏼‍♂️

PR #484, opened 2025-03-20
Test: Only attach "closePopupsIfExist" once per test

marusak:
Having generic close functions can close pop-ups that we don't count with and cause unexpected behavior.

**Testing** what popups we still need to close.

### DIFF
--- a/_playwright-tests/UI/Templates.spec.ts
+++ b/_playwright-tests/UI/Templates.spec.ts
@@ -33,7 +33,6 @@ test.describe('Templates', () => {
 
     await test.step('Navigate to the templates page', async () => {
       await navigateToTemplates(page);
-      await closePopupsIfExist(page);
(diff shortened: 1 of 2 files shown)


dominikvagner:
I don't think there will be any locator handlers in the tests where you removed those calls to the `closePopupsIfExist` function. 😕 The `Page` objects are isolated per test so the handlers won't 'survive' to the next one. 

If that is the desired outcome (i.e.: to remove those handler from those specific tests) then sure, let's merge it, but given this PRs title, I don't think that is it (or is it just misleading?), because those tests won't have those handlers at all, not 'attached once'.

marusak:
Thanks, they I misunderstood how it works. I saw weird behaviour where one popup triggered twice the handling action but I guess it was a different problem then. Closing!

PR #486, opened 2025-03-25
HMS-5678: check for use latest on warning in the set up date step

dominikvagner:
## Summary
This fixes a bug where the warning about snapshots before the selected date was displayed in the set up date step inside of the template add modal even when use latest was selected.

## Testing steps
When creating or editing a template the warning should not be displayed if the use latest option is selected in the set up date step.


jlsherrill:
https://issues.redhat.com/browse/HMS-5678

marusak (approved):
Thanks!

PR #495, opened 2025-04-04
HMS-5813: QE: migrate Template CRUD test

swadeley:
## Summary

[HMS-5813] QE: migrate Template CRUD test

## Testing steps

peer review

PR #499, opened 2025-04-09
Build: add centirepos

swadeley:
## Summary

I made a new batch of 100 repos without a number in the URL
The multirepos path with a rank number in URL causes problems when that number corresponds to a release version.

## Testing steps

Tests pass


_playwright-tests/UI/CustomRepoCRUD.spec.ts
```diff
@@ -2,17 +2,12 @@ import { test, expect } from '@playwright/test';
...
-export const rank = () => Math.floor(Math.random() * 10 + 1).toString();
-export const randomNum = () =>
-  Math.floor(Math.random() * 10 + 1)
-    .toString()
-    .padStart(2, '0');
+const repoNamePrefix = 'Repo-CRUD';
+const repoName = `${repoNamePrefix}-${randomName()}`;
 
-export const url = `https://stephenw.fedorapeople.org/multirepos/${rank()}/repo${randomNum()}/`;
+const url = `${randomUrl()}`;
```
dominikvagner, line 10: why not just `const url = randomUrl()`, no need to wrap it in a format string 😅

_playwright-tests/UI/helpers/repoHelpers.ts
```diff
@@ -0,0 +1,8 @@
+export const randomName = () => (Math.random() + 1).toString(36).substring(2, 6);
+
+export const randomNum = () =>
+  Math.floor(Math.random() * 100 + 1)
+    .toString()
+    .padStart(2, '0');
+
+export const randomUrl = () => `https://stephenw.fedorapeople.org/centirepos/repo${randomNum()}/`;
```
dominikvagner, line 8: this is nice, much simpler this way 👍🏼 

but it would be even better to have those repos in our [fixtures repo](https://github.com/content-services/fixtures/), which was created for this exact purpose, i.e.: hosting our's testing repositories, where the ownership would be on the entire team 😄

dominikvagner (commented):
left two comments 💭😅

_playwright-tests/UI/CustomRepoCRUD.spec.ts
swadeley: yes, good point, fixed, thank you

_playwright-tests/UI/helpers/repoHelpers.ts
swadeley: RFE versus bugfix

swadeley: Hi, I think this might work:

https://github.com/content-services/fixtures/pull/6

dominikvagner: that would, awesome! 👏🏼 
just the URL needs to be changed now to point to the new location 😄

swadeley: @dominikvagner rebased and updated to use github repo path. 
Thank you.

swadeley: Changes to repo path also applied in:

https://github.com/content-services/content-sources-frontend/pull/495

swadeley: If you merge https://github.com/content-services/content-sources-frontend/pull/495 we can close this PR.

PR #501, opened 2025-04-10
Test: Wait longer for the page to load

marusak:
Do not merge, just trying stuff out!


marusak:
GET on fed-modules.json is repeating pattern :thinking:

PR #502, opened 2025-04-11
Test: merge incorrectly split tests, improve speed

dominikvagner:
## Summary
This merges test that were incorrectly split into multiple and substitutes the separation with more steps. 🧹  
Also make a small change to navigation helpers to hopefully make them a little faster. ⚡ 

## Testing steps
Tests pass and hopefully don't time out in CI + should be less flaky due to decreased number of navigations and increased retries 🙏🏼


swadeley (approved):
ACK

PR #503, opened 2025-04-11
HMS-5940: Add validation for creating content templates in Templates.spec.ts

mayurilahane:


mayurilahane:
We can use the API method introduced in this PR for template creation.

however, it is not a priority at the moment, so I am closing it. 

Regarding the feature to verify the valid status of templates, it is addressed as part of PR #495

PR #512, opened 2025-04-22
HMS-5508: Patternfly 5 - 6 update

Andrewgdewar:
## Summary

- Updates Patternfly and associated packages to pf6
- Touches a lot of files 😞


jlsherrill:
https://issues.redhat.com/browse/HMS-5508

rverdile (commented):
I've been using this as my frontend to see if I notice any problems, and for the most part I don't. I've noticed only two things, but these could also just be issues with patternfly.

1) The repositories and templates are kinda smushed on my screen. I can't see the column headers, the "18 hours ago" text needs to two lines,  the status needs two lines.
**Before:**
image

**After:**
image

2) the "In progress" status text is kinda purple 

Added a before screenshot. And actually it was also an issue on patternfly 5, although it is a little worse now

xbhouse:
> I've been using this as my frontend to see if I notice any problems, and for the most part I don't. I've noticed only two things, but these could also just be issues with patternfly.
> 
>     1. The repositories and templates are kinda smushed on my screen. I can't see the column headers, the "18 hours ago" text needs to two lines,  the status needs two lines.
>        **Before:**
>        image
> 
> 
> **After:** image
> 
>     2. the "In progress" status text is kinda purple
> 
> 
> Edit: added a before screenshot. And actually it was also an issue on patternfly 5, although it is a little worse now

pushed a fix to try to match the colors as much as possible to what we have currently, there were lots of pastels by default :laughing: 

and adjusted the padding on the table, it's still not great but maybe not worse than it is in pf5? let me know how it's looking on your end

rverdile:
looks better!

_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -35,15 +35,14 @@ test.describe('Snapshot Repositories', () => {
 
     await test.step('Filter by architecture', async () => {
       await page.getByRole('button', { name: 'filter architecture' }).click();
-      await page.getByRole('option', { name: 'x86_64' }).click();
+      await page.getByTestId('filter_x86_64').click();
```
dominikvagner, line 38: if this is only possible to target with testid than that sucks big time, I wonder is it a configuration issue on our side or if it really coming from patternfly as I am getting this error in the console on clicking any dropdown options/items 🤔 
we should investigate this, if not in this PR, then in a follow up 🔍 

2025-05-21-150542_hyprshot

dominikvagner (commented):
added few comments, but overall looks good! (_the PR not the new Patternfly, that's ugly_ 🤢😂)
some of them could/should be solved as followups so this isn't blocked much further and we can merge it sooner before it deviates a lot 😅

also this needs to be rebased, because of conflicts and to get the playwright CI fixes so the tests have some chance of passing 😄

src/Pages/Repositories/ContentListTable/ContentListTable.tsx
```diff
@@ -93,6 +90,11 @@ const useStyles = createUseStyles({
   uploadIcon: {
     marginLeft: '8px',
   },
+  tightPadding: {
+    '& th, & td': {
+      padding: '4px 8px',
+    },
+  },
```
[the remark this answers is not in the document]
xbhouse: makes sense :laughing: :cry:  removed the custom css, i can open a ticket with the PF team to try and improve this

_playwright-tests/UI/TemplateCRUD.spec.ts
```diff
@@ -34,16 +34,16 @@ test.describe('Templates CRUD', () => {
     });
     await test.step('Create a template', async () => {
       await page.getByRole('button', { name: 'Add content template' }).click();
-      await page.getByRole('button', { name: 'Select architecture' }).click();
-      await page.getByRole('option', { name: 'aarch64' }).click();
-      await page.getByRole('button', { name: 'Select version' }).click();
-      await page.getByRole('option', { name: 'el9' }).click();
+      await page.getByRole('button', { name: 'filter arch' }).click();
+      await page.getByTestId('filter_aarch64').click();
```
xbhouse, line 38: changed these to use test ID since that's the only thing that seems to work so far, i remember Andrew was fighting with the dropdown targeting :cry: still investigating

dominikvagner: this seems to do the job 😅 _(pro tip: every time a locator is weird try launching the playwright debug mode `--debug` and using the pick locator tool they have, also you live edit it in the bottom split to tweak it 🤓)_
```ts
await page.getByRole('menuitem', { name: 'aarch64' }).click();
```

xbhouse: hmmm i tried using menuitem and it didn't work :thinking: let me try again

xbhouse: ok! that does work for the add template modal, but not for the add repo modal :laughing: but it _looks like_ menuitem should work for both:

add repo modal: `<button tabindex="-1" class="pf-v6-c-menu__item" role="menuitem" type="button"><span class="pf-v6-c-menu__item-main"><span class="pf-v6-c-menu__item-text">x86_64</span></span></button>` 

add template modal: `<button tabindex="0" class="pf-v6-c-menu__item" role="menuitem" type="button"><span class="pf-v6-c-menu__item-main"><span class="pf-v6-c-menu__item-text">aarch64</span></span></button>`

xbhouse: i think (and i may be wrong, but it's the only difference i see) for some reason getbyrole isn't working with the repo add modal dropdowns because aria-hidden is true on those menus, but not on the template repo add modal dropdowns :exploding_head: 

we can work around using getbyrole by doing something like i have currently pushed up or maybe this: `await page.locator('[role="menu"] [role="menuitem"]', { hasText: 'x86_64' }).click();` and i can open a ticket for this as a follow-up? that way we can get the majority of these changes checked out by UX sooner

_playwright-tests/UI/SnapshotRepo.spec.ts
xbhouse: opened this for removing the deprecated components: https://issues.redhat.com/browse/HMS-8602

_playwright-tests/UI/TemplateCRUD.spec.ts
dominikvagner: yeah, you're right, the `aria-hidden` set on that menu breaks Playwright as it isn't available through the accessibility API when it's aria hidden 😒 
spent couple hours looking for a fix and didn't find anything which would solve this 🥲 also looked in the PF docs and even their "modal with dropdown" [example](https://www.patternfly.org/components/modal#with-dropdown) is broken (i.e.: throws this error about blocking aria-hidden on an element)❗  
2025-05-26-143630_hyprshot

this is very annoying, as this makes the modal unusable for anyone using assistive technologies, proper testing and the form is also not able to be fully filled only using a keyboard, because unless you click any of the menu items first prompting the browser to block the aria hidden attribute you can't select the arch or dist with just using the keyboard 😞

but as this doesn't seem to be the fault of this PR (even though I don't understand why it works in the template modal/wizard 😵‍💫) I also think we should create a follow-up for this, where part of it would be filing a bug in PF to fix this 😅

xbhouse: started a thread with the patternfly team [here](https://redhat-internal.slack.com/archives/C04JMHKSD9C/p1748354607400429)

dominikvagner (approved):
re-reviewed and this seems to be good to go! 🙌🏼 all tests are passing (konflux+playwright) 🚀 
let's not wait for long before merging this, approving! ✅ _(although it might be good to have more than 1 ack, becasue of the size of this PR 😅)_

awesome job @Andrewgdewar and @xbhouse for getting this over finish line! 👏🏼🫶

deploy/frontend.yaml
```diff
@@ -18,15 +18,15 @@ objects:
...
-      navItems:
-        - appId: "content-sources"
-          title: "Repositories"
-          href: "/insights/content/repositories"
-          product: "Content"
-        - appId: "content-sources"
-          title: "Templates"
-          href: "/insights/content/templates"
-          product: "Content"
+      # navItems:
```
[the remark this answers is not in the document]
xbhouse: with this version of fec, when the navItems are left in we see:

```
  Frontend CRD validation error:
    - must NOT have additional properties
      /objects/0/spec
      additionalProperties
      {"additionalProperty":"navItems"}
[fec] Warn:  Error: Frontend CRD validation failed! must NOT have additional properties
```

i wonder if those are unnecessary if we have the module property? Andrew's comment on this change notes `fec update` so it could be related to the fec version change

src/Pages/Repositories/RepositoryLayout.tsx
```diff
@@ -33,7 +29,7 @@ const useStyles = createUseStyles({
     },
   },
   containerMargin: {
-    margin: '24px',
+    margin: '16px 24px', // TODO: Check if side padding is desired
```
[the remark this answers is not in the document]
rverdile: I also don't really see a difference

rverdile:
this looks good to me

xbhouse:
thanks for all your work @Andrewgdewar :heart: merging to get these changes in

PR #516, opened 2025-04-25
Change pf-v5 to pf-v6 in locators

swadeley:
## Summary

Stage is now pf-v6

## Testing steps

tests pass

NOTE: We should not be using this type of locator.


Andrewgdewar:
Already updated in the pf6 upgrade

PR #524, opened 2025-05-02
QE - migrate iqe test_snapshot_deletion

mayurilahane:


_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -103,3 +103,117 @@ test.describe('Snapshot Repositories', () => {
...
   });
 });
+
+test('Snapshot deletion', async ({ page }) => {
+  await navigateToRepositories(page);
+  await closePopupsIfExist(page);
+  const repoNamePrefix = 'snapshot-deletion';
+  const randomName = () => `${(Math.random() + 1).toString(36).substring(2, 6)}`;
+  const repoName = `${repoNamePrefix}-${randomName()}`;
+  const templateName = `Test-template-for-snapshot-deletion-${randomName()}`;
```
dominikvagner, line 113: you should also add a cleanup of repos at the start here, and since you are using specific URLs, not a random one, you need to delete any repos that use any of the zoo repos 🧹

_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -103,3 +103,117 @@ test.describe('Snapshot Repositories', () => {
...
+    await page.getByPlaceholder('Description').fill('Template test');
+    await page.getByRole('button', { name: 'Next', exact: true }).click();
+    await page.getByRole('button', { name: 'Create other options' }).click();
+    await page.getByText('Create template only', { exact: true }).click();
+    const templateRow = await getRowByNameOrUrl(page, templateName);
+    await expect(templateRow.getByText('Valid')).toBeVisible({ timeout: 60000 });
+    // Verify the template is created and uses the latest snapshot
+    await expect(
+      await page.getByRole('gridcell', { name: 'Use latest' }).first().textContent(),
+    ).toBe('Use latest');
```
dominikvagner, line 175: this expect is a bit weird since it gets a cell with 'use latest' text and verifies it has that text (just asserting that text to be visible would be a bit more clear) 🤔 
also this would pass if there was any template with the use latest, it doesn't check the created template, needs to target something more specific 🎯

_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -103,3 +103,117 @@ test.describe('Snapshot Repositories', () => {
...
+    // Assert that the template snapshot count matches the repo snapshot count
+    expect(snapshotCount).toBe(4);
+
+    // Test deletion of a single snapshot.
+    await test.step('Delete a single snapshot', async () => {
```
dominikvagner, line 180: this step (and the bulk delete one) is nested inside of another one (`Verify snapshot count for the repo`), it doesn't really make sense for it to be there from the test point of view and in general I would not nest any steps 😅

_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -103,3 +103,117 @@ test.describe('Snapshot Repositories', () => {
...
+  await test.step('Verify the snapshot count for the repo.', async () => {
+    const row = await getRowByNameOrUrl(page, repoName);
+    await expect(row.getByText('Valid')).toBeVisible({ timeout: 60000 });
+    await page.getByRole('button', { name: 'Kebab toggle' }).click();
+    await page.getByRole('menuitem', { name: 'View all snapshots' }).click();
+    // Count the number of rows in the snapshot list table
+    const snapshotCount = (await page.getByTestId('snapshot_list_table').locator('tr').count()) - 1; // Subtract 1 for the header row
+    // Create a template which uses the repo and assert that is uses the latest snapshot
+    await page.getByLabel('Close', { exact: true }).click();
+    await navigateToTemplates(page);
```
dominikvagner, line 152: the template creation could be a separate step 💅🏼 (i.e.: cut the verify count step after line 151 and make the template creation a separate step)

dominikvagner (changes requested):
added a few comments, some changes needed 😅

_unable to run the test currently due to a bug on stage, will revisit after that is fixed and comments addressed 🫡_

_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -103,3 +103,117 @@ test.describe('Snapshot Repositories', () => {
...
+    await page.getByRole('button', { name: 'Add content template' }).click();
+    await page.getByRole('button', { name: 'Select architecture' }).click();
+    await page.getByRole('option', { name: 'aarch64' }).click();
+    await page.getByRole('button', { name: 'Select version' }).click();
+    await page.getByRole('option', { name: 'el9' }).click();
+    await page.getByRole('button', { name: 'Next', exact: true }).click();
+    await page.getByRole('button', { name: 'Next', exact: true }).click();
+    await row.getByRole('gridcell', { name: 'Select row' }).locator('label').click();
+    await page.getByRole('button', { name: 'Next', exact: true }).click();
+    await page.getByText('Use latest contentAlways use').click();
```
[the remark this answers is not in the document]
mayurilahane: good catch ! badly copy pasted 

THanks !

_playwright-tests/UI/SnapshotRepo.spec.ts
mayurilahane: I understood what are you suggesting made the changes and seperated it in different test step now 
thanks !

_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -103,3 +103,117 @@ test.describe('Snapshot Repositories', () => {
     });
   });
 });
+
+test('Snapshot deletion', async ({ page }) => {
```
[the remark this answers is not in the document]
dominikvagner: this isn't resolved, you added a new describe block instead moving the test into the existing one, please fix that 🔧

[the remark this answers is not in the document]
mayurilahane: Done

_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -102,4 +102,130 @@ test.describe('Snapshot Repositories', () => {
...
+        (await page.getByTestId('snapshot_list_table').locator('tr').count()) - 1; // Subtract 1 for the header row
+      expect(snapshotCount).toBe(4);
+      await navigateToTemplates(page);
+      await page.getByRole('button', { name: 'Add content template' }).click();
+      await page.getByRole('button', { name: 'Select architecture' }).click();
+      await page.getByRole('option', { name: 'aarch64' }).click();
+      await page.getByRole('button', { name: 'Select version' }).click();
+      await page.getByRole('option', { name: 'el9' }).click();
+      await page.getByRole('button', { name: 'Next', exact: true }).click();
+      await page.getByRole('button', { name: 'Next', exact: true }).click();
```
marusak, line 171: This failed in CI. I did not investigate a lot, but my first thought was - You click on a button and then want to click on it on the next screen. But the test is quick, so it will probably click on the same button twice - before it gets to the other screen. You should put there some wait for the next screen to be visible before clicking next time.

_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -102,4 +102,130 @@ test.describe('Snapshot Repositories', () => {
...
+      await page
+        .getByTestId('snapshot_list_table')
+        .locator('tbody tr')
+        .first()
+        .getByLabel('Kebab toggle')
+        .click();
+      await page.getByRole('menuitem', { name: 'Delete' }).click();
+      await expect(page.getByText('Remove snapshots?')).toBeVisible();
+      await page.getByText('Remove', { exact: true }).click();
+      await page.waitForTimeout(2000);
```
marusak, line 204: Please don't use sit!

_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -102,4 +102,130 @@ test.describe('Snapshot Repositories', () => {
...
+      await page.getByRole('checkbox', { name: 'Select row 0' }).uncheck();
+      await page.getByTestId('remove_snapshots_bulk').click();
+      await expect(page.getByText('Remove snapshots?')).toBeVisible();
+      await page.getByText('Remove', { exact: true }).click();
+      await page.waitForTimeout(2000);
+      // Verify that the remaining snapshot count is 1 after bulk deletion
+      const snapshotCountAfterBulkDeletion =
+        (await page.getByTestId('snapshot_list_table').locator('tr').count()) - 1; // Subtract 1 for the header row
+      expect(snapshotCountAfterBulkDeletion).toBe(1);
+      await page.getByText('Close').click();
```
marusak, line 227: We talked about this what would be a correct solution. Why did you decide to just sleep for 2 seconds?

_playwright-tests/UI/SnapshotRepo.spec.ts
mayurilahane: This approach is working well for me now.

But i think you're right, it could lead to issues in the future.
So I'm modifying this line and adding an assertion to ensure the model is present for both this and the next page models.

Thanks!

_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -102,4 +102,130 @@ test.describe('Snapshot Repositories', () => {
...
+        .click();
+      await page.getByRole('menuitem', { name: 'Delete' }).click();
+      await expect(page.getByText('Remove snapshots?')).toBeVisible();
+      await page.getByText('Remove', { exact: true }).click();
+      await page.waitForTimeout(2000);
+
+      const snapshotCountAfterSingleDeletion =
+        (await page.getByTestId('snapshot_list_table').locator('tr').count()) - 1; // Subtract 1 for the header row
+      expect(snapshotCountAfterSingleDeletion).toBe(3);
+    });
```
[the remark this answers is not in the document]
mayurilahane: Yes, I had implemented it earlier as well, but I unnecessarily overcomplicated it. 
[here](https://github.com/content-services/content-sources-frontend/pull/524/commits/68c232f6128afcab39176de5eaca0f05ce177531#diff-9b23ee601442d5a43ba5875e94378668b283e911eae974f16f2c3babdcd1a1e6R162)  

But, I identified the issue and have now corrected it.

_playwright-tests/UI/SnapshotRepo.spec.ts
mayurilahane: Yes removed it.
Thanks !

_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -102,4 +102,154 @@ test.describe('Snapshot Repositories', () => {
...
+        .replace(',', ''); // Format: 20 May 2025
+      const targetRows = page.getByRole('row').filter({ has: page.getByText(snapshotPrefix) });
+      await expect(targetRows).toHaveCount(3);
+      await page.getByText('Close').click();
+    });
+
+    await test.step('Bulk delete snapshot', async () => {
+      // Test bulk deletion of multiple snapshots.
+      // Before bulk delete we need to wait for previous deletion to finish or it will fail.
+      await page.waitForTimeout(6000);
```
mayurilahane, line 225: We need to add a timeout here, as the bulk delete operation fails by mistakenly recognizing it as the same delete action from the previous single delete test step.

dominikvagner (changes requested):
few comments, some possible improvements and a definite bug in this (the one Matej pointed out) 😅 
now that our CI is stable/working, nothing on which Playwright tests are failing should be merged 🚫

_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -102,4 +102,157 @@ test.describe('Snapshot Repositories', () => {
...
+      await page.getByRole('menuitem', { name: 'View all snapshots' }).click();
+      const snapshotPrefix = new Date()
+        .toLocaleString('en-GB', {
+          day: '2-digit',
+          month: 'short',
+          year: 'numeric',
+          hour12: false,
+        })
+        .replace(',', ''); // Format: 20 May 2025
+      const targetRows = page.getByRole('row').filter({ has: page.getByText(snapshotPrefix) });
```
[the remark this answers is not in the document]
dominikvagner: yeah, that seems like a nicer solution but I know that was giving you trouble before 😅 just checking that the pagination, shows the correct number of snapshots would be completely fine and simpler, i.e.: 
```ts
await expect(getByRole('button', { name: '1 - 1 of 1' }).first()).toBeVisible() // replace the numbers for the correct/expected amount of snapshots 
```

_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -102,4 +102,157 @@ test.describe('Snapshot Repositories', () => {
...
+
+    await test.step('Cleanup repositories using "zoo" URLs', async () => {
+      const zooUrls = [
+        'https://fedorapeople.org/groups/katello/fakerepos/zoo',
+        'https://fedorapeople.org/groups/katello/fakerepos/zoo2',
+        'https://fedorapeople.org/groups/katello/fakerepos/zoo3',
+        'https://fedorapeople.org/groups/katello/fakerepos/zoo4',
+      ];
+      for (const url of zooUrls) {
+        await deleteAllRepos(page, `&url=${url}`);
```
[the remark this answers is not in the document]
dominikvagner: ehm, what? 🤔 not sure I follow you here 😅 
this has nothing to do with any UI and dynamic text in buttons, this is a cleanup using the API.

no need to add any bulk delete method, the delete all repos helper does that, but in the way you are using it you are unnecessarily calling it 4 times instead of one sufficient call.
the API for listing repositories _(the one you are filtering with the second argument)_ can accept multiple URL joined by a comma so if you'd joined them you can only call it once 💭😄 
or just call it once with this URL `https://fedorapeople.org/groups/katello/fakerepos/zoo`, without the backslash at the end, and that will return all repos starting with that string, so the ones that you want to remove 💁🏼‍♂️

_playwright-tests/UI/SnapshotRepo.spec.ts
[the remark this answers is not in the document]
mayurilahane: tried with this locator but its not working 

``` 
await expect(page.getByRole('button', { name: '1-3 of 3' }).first()).toBeVisible({
        timeout: 60000,
      });
```

then tried to choose another way to locate the whole pagination widget and then read its text content of it 
like this :

```
const pagination = model.locator('#top-pagination-id').first().textContent();
```
but it returns an whole object and not the text content

also tried with defining modal and then using the pagination widget 
like this 
```
 const model = page.getByTestId('snapshot_list_modal');
      await model.locator('span.pf-v5-c-menu-toggle__text').first().isVisible();
      const pagination = model.locator('span.pf-v5-c-menu-toggle__text').first().textContent();
      await expect(pagination).toContain('1-3 of 3');
``` 

but again it is returning an object 

```
Expected value:  "1-3 of 3"
Received object: {Symbol(async_id_symbol): 10343, Symbol(trigger_async_id_symbol): 10228, Symbol(kResourceStore): {"_asyncLocalStorage": {"enabled": true, "kResourceStore": Symbol(kResourceStore)}, "_data": Map {"stepZone" => {"attachmentIndices": [], "box": undefined, "boxedStack": undefined, "category": "test.step", "complete": [Function complete], "location": {"column": 16, "file": "_playwright-tests/UI/SnapshotRepo.spec.ts", "function": undefined, "line": 198}, "stepId": "test.step@113", "steps": [{"apiName": "page.route", "attachmentIndices": [], "boxedStack": undefined, "category": "pw:api", "complete": [Function complete], "endWallTime": 1748462697675, "location": {"column": 16, "file": "/home/mlahane/satellite/content-sources-frontend/_playwright
```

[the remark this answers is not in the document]
mayurilahane: Added suggested changes thanks !

_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -101,4 +101,132 @@ test.describe('Snapshot Repositories', () => {
...
+    });
+
+    await test.step('Edit the repository', async () => {
+      for (let i = 2; i <= 4; i++) {
+        const row = await getRowByNameOrUrl(page, repoName);
+        await expect(row.getByText('Valid')).toBeVisible({ timeout: 60000 });
+        await test.step(`Edit repository and create snapshot ${i}`, async () => {
+          // Open the edit modal
+          await row.getByLabel('Kebab toggle').click();
+          await row.getByRole('menuitem', { name: 'Edit' }).click({ timeout: 60000 });
```
xbhouse, line 137: the dropdown items are located only at the page level in PF6, so to target the Edit menuitem we'll need to change this to `await page.getByRole('menuitem', { name: 'Edit' }).click({ timeout: 60000 });`

PR #527, opened 2025-05-13
HMS-8563: change to testing proxy for PW tests in actions

dominikvagner:
## Summary
This PR changes the playwright action to use a new custom testing proxy, which should make tests faster and more stable.
For the custom proxy a new config file with routes was added.
Also resets the sharding of PW tests as in our case one larger worker seems to be a better choice.
And fixes some minor issues in our tests (awaits,timeouts, etc.).

## Testing steps
Playwright test should pass without loading errors.


marusak (approved):
Nice! :heart:

jlsherrill:
https://issues.redhat.com/browse/HMS-8563

PR #528, opened 2025-05-15
Fix: Add breadcrumb to navlink locator

swadeley:
## Summary

Without this change the locator can resolve to two locations of "Repositories", one in breadcrumbs and one in side bar. We just want the first one to be resolved in the breadcrumbs trail as a sign the page is loaded and the correct page.

## Testing steps
 tests pass


dominikvagner (approved):
looks good, tests are green! 🎉 nice job 👏🏼 
approved ✅ 

_the bit unrelated timeout increase is fine here, since the navigation with the new testing proxy doesn't need to retry loading the page several times so this won't be much of a problem/duration increase 😅_

PR #529, opened 2025-05-15
Refactor logout helper

swadeley:
## Summary

Current logout helper relies on PatternFly-specific classes to find the menu.
Also, the menu sometimes closes during manual testing. In this PR I do not seek to add a retry block, but if it closes we will get more useful error messages.

The change does require passing the user name as the menu dropdown button looks like this:
<span class="pf-v6-c-menu-toggle__text">HMS QA</span>

## Testing steps

I will test with :
[HMS-5999: QE: migrate iqe test_rbac_for_repos
](https://github.com/content-services/content-sources-frontend)

### DIFF
--- a/_playwright-tests/helpers/loginHelpers.tsx
+++ b/_playwright-tests/helpers/loginHelpers.tsx
@@ -3,23 +3,29 @@ import path from 'path';
 
 // This file can only contain functions that are referenced by authentication.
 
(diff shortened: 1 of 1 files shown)


swadeley:
hi

this approch is not good because the dropdown menu name displayed is "First name last name", `username` is only displayed when you open the menu.

PR #538, opened 2025-05-27
HMS-6036: QE - Enhancing current test case for "Popular repositories"

mayurilahane:
Migrate IQE 

- test_pop_add_all
- test_pop_add_all_introspection
- test_add_one_popular_repo


_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -22,52 +22,29 @@ test.describe('Popular Repositories', () => {
       await expect(page.getByRole('button', { name: 'Add selected repositories' })).toBeVisible();
```
xbhouse, line 22: as part of this, do you mind renaming this test from `Add popular repos` to `Test adding and removing popular repos` since it's doing more than just adding them?

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -22,52 +22,29 @@ test.describe('Popular Repositories', () => {
...
-
-    await test.step('Select EPEL 8 and add both repos', async () => {
-      await page
-        .getByRole('row', { name: repoName8 })
-        .getByLabel('Select row 2', { exact: true })
-        .click();
-      await page.getByTestId('add-selected-dropdown-toggle-no-snap').click();
-      await page.getByRole('menuitem', { name: 'Add 3 repositories without snapshotting' }).click();
-    });
+    // Select the all 3 popular repositories and add them without snapshotting.
```
xbhouse, line 25: this block (lines 26-36) can be included in its own test.step, maybe something like `await test.step('Add the 3 popular repos without snapshotting', async () => {...});`. then you can remove the `await test.step(Select ${repos[i]}` inside the for loop, and probably the comment at line 25 too

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -22,52 +22,28 @@ test.describe('Popular Repositories', () => {
...
-      await page.getByRole('menuitem', { name: 'Add 2 repositories without snapshotting' });
-    });
-
-    await test.step('Select EPEL 8 and add both repos', async () => {
-      await page
-        .getByRole('row', { name: repoName8 })
-        .getByLabel('Select row 2', { exact: true })
-        .click();
+    await test.step('Add the 3 popular repos without snapshotting', async () => {
+      const repos = [repoName10, repoName9, repoName8];
```
dominikvagner, line 26: you are defining this 4 times at different places, instead this should be defined once globally so it can be reused ♻️

dominikvagner (commented):
one comment with suggested improvement and one nitpick 😄 
otherwise looking great. makes the popular repos test nicer ✨👍🏼

PR #545, opened 2025-06-14
HMS-5828: add shared epel and remove popular repos

xbhouse:
## Summary

* Dependent on [this backend PR](https://github.com/content-services/content-sources-backend/pull/1117)
* Adds an EPEL toggle and changes toggle behavior to be multi-select for Custom, EPEL, and Red Hat repositories
* Visibility of the Popular repositories tab and EPEL toggle is controlled with the [community repos feature](https://github.com/content-services/content-sources-backend/blob/main/configs/config.yaml.example#L173)

## Testing steps

1. Set the [community repos feature](https://github.com/content-services/content-sources-backend/blob/main/configs/config.yaml.example#L173) to true
2. Run `make repos-import && ./release/external-repos process-repos`
3. Verify selecting/deselecting the Custom, EPEL, and Red Hat toggles shows the expected content
4. Verify the shared EPEL and Red Hat repos cannot be edited or deleted
5. When navigating to the repositories page from the zero state, the Custom and EPEL toggles should be selected by default 
6. Deleting repositories should work correctly (confirm this works regardless of which number page you're on)
7. Set the community repos feature to false. The Popular repositories tab should still be visible, the EPEL toggle should not be, and the zero state should still have a route to the Popular repositories tab

### DIFF
--- a/.github/workflows/playwright-actions.yml
+++ b/.github/workflows/playwright-actions.yml
@@ -196,6 +196,10 @@ jobs:
         working-directory: content-sources-backend
         run: OPTIONS_REPOSITORY_IMPORT_FILTER=small make repos-import
 
+      - name: Import the shared EPEL10 repo
+        working-directory: content-sources-backend
+        run: OPTIONS_REPOSITORY_IMPORT_FILTER=epel10 make repos-import
+
       - name: Run snapshot for redhat repo
         working-directory: content-sources-backend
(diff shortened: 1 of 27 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-5828

rverdile (commented):
tested, looks good

rverdile (commented):
I think I'm seeing one issue where there's no zero state if community repos are enabled. If I have no custom repositories, and navigate to the repositories page, it just loads the dual-selected `Custom | EPEL` list without me having to click through the zero state first. Do see that as well?

Everything else looks good!

PR #549, opened 2025-06-18
Remove "first()" to enable multiple popup closing

swadeley:
## Summary

We should not use `fits()` in this context with `noWaitAfter: true`

It was failing to close multiple popups which matched `button[id^="pendo-close-guide-"`

## Testing steps
Tests pass


marusak:
Won't it fail if `locator` resolves to multiple elements?

dominikvagner:
> Won't it fail if `locator` resolves to multiple elements?

yeah, I am also worried about this 🤔 
the closing of pop-ups is made in way so that it doesn't crash the test (the surrounding try/catch throws away those errors on purpose), but if there are multiple of similar popups, each of the handlers would just fail and not close any one of them I think 💭 

@swadeley do you have examples of the problem this fixes (links to recordings/traces)? and have you verified this wouldn't break other scenarios? 👀

PR #550, opened 2025-06-19
Test: Run slack notification even when test fails

marusak:
## Summary by Sourcery

CI:
- Refine Slack Notification condition to trigger for non-pull_request events unless the job is cancelled


dominikvagner (approved):
thanks, approved! ✅

PR #555, opened 2025-06-25
Fix(CI): change nightly tests to checkout submodules

dominikvagner:
## Summary
This adds missing submodule checkout to the nightly-stage-test-action.


.github/workflows/playwright-actions.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       - name: Create front-end .env file
         run: |
           # Check if the secrets are set
-          if [ -z "$" ]; then
+          if [ -z "$ADMIN_USERNAME" ]; then
```
marusak, line 31: ? Seems obvious, but some mention in commit message would be nice :)

marusak (approved):

.github/workflows/playwright-actions.yml
dominikvagner: thought it was obvious, but amended it to the commit message 🫡😅

PR #558, opened 2025-07-02
HMS-8739: add integration test for updating system with template

rverdile:
## Summary
Adds an integration test which

1. Creates a template with an old snapshot
2. Registers that template to a system
3. Fetches the newest advisory on that system
4. Changes the template date to the latest snapshot
5. Refreshes the system to get the new content
6. Fetches the newest advisory
7. Verifies the advisory has changed


_playwright-tests/Integration/CanUpdateSystemWithTemplate.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+        'dnf updateinfo --list --all | grep RH | sort | tail -n 1',
+      ]);
+      if (exist?.exitCode != 0) {
+        console.log(exist?.stdout);
+        console.log(exist?.stderr);
+      }
+      expect(exist?.exitCode).toBe(0);
+      const secondVersion = exist?.stdout?.toString();
+      expect(secondVersion).not.toBe(firstVersion);
+    });
```
jlsherrill, line 139: Would you also be able to add a test that installs a package (something like vim-enhanced?)  This would verify that RPMs can be downloaded too

jlsherrill: and it wouldn't really matter which version of the template is used

jlsherrill:
overall looks good.  the test run failed with ```ERROR  rhsm  cannot connect to Red Hat Subscription Management: error: Organization 17684638 does not exist.```

_playwright-tests/Integration/CanUpdateSystemWithTemplate.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+      expect(exist?.exitCode).toBe(0);
+      firstVersion = exist?.stdout?.toString();
+    });
+
+    await test.step('Update the template date to latest', async () => {
+      const rowTemplate = await getRowByNameOrUrl(page, templateName);
+      await rowTemplate.getByRole('button', { name: templateName }).click();
+      await expect(page.getByLabel('Breadcrumb').first()).toHaveText('RHELContentTemplates');
+      await expect(page.getByRole('heading', { level: 1 })).toHaveText(templateName);
+      await expect(page.getByText('Description:Template test')).toBeVisible();
```
marusak, line 92: What do these 3 rows do? You are waiting for the page to load?

_playwright-tests/Integration/CanUpdateSystemWithTemplate.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+        page.getByRole('heading', { name: 'Define template content', exact: true }),
+      ).toBeVisible();
+      await page.getByRole('button', { name: 'Next', exact: true }).click();
+      await expect(
+        page.getByRole('heading', { name: 'Additional Red Hat repositories', exact: true }),
+      ).toBeVisible();
+      await page.getByRole('button', { name: 'Next', exact: true }).click();
+      await expect(
+        page.getByRole('heading', { name: 'Custom repositories', exact: true }),
+      ).toBeVisible();
```
marusak, line 105: This is pretty much what I commented above about those 3 clicks on Next.

_playwright-tests/Integration/CanUpdateSystemWithTemplate.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+      await page.getByRole('button', { name: 'Add content template' }).click();
+      await page.getByRole('button', { name: 'filter arch' }).click();
+      await page.getByRole('menuitem', { name: 'x86_64' }).click();
+      await page.getByRole('button', { name: 'filter version' }).click();
+      await page.getByRole('menuitem', { name: 'el9' }).click();
+      await page.getByRole('button', { name: 'Next', exact: true }).click();
+      await page.getByRole('button', { name: 'Next', exact: true }).click();
+      await page.getByRole('button', { name: 'Next', exact: true }).click();
+      await page.getByText('Use up to specific date', { exact: true }).click();
+      await page.getByPlaceholder('YYYY-MM-DD', { exact: true }).fill('2021-05-17'); // Older than any snapshot date
```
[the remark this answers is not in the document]
rverdile: It's older than any possible snapshot so it will select the oldest snapshot

_playwright-tests/Integration/CanUpdateSystemWithTemplate.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+      await page.getByRole('button', { name: 'Next', exact: true }).click();
+      await page.getByRole('button', { name: 'Create other options' }).click();
+      await page.getByText('Create template only', { exact: true }).click();
+      const rowTemplate = await getRowByNameOrUrl(page, `${templateName}`);
+      await expect(rowTemplate.getByText('Valid')).toBeVisible({ timeout: 60000 });
+    });
+
+    await test.step('Create RHSM client and register the template', async () => {
+      // change the test timeout as registering a client can be slow
+      test.setTimeout(10 * 60 * 1000); // ten minutes
```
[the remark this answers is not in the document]
rverdile: This is global. I'd have preferred to scope it to the step, but I didn't see a way to do that. However, I removed this in favor of putting the timeout directly on the commands I saw take a while.

_playwright-tests/Integration/CanUpdateSystemWithTemplate.spec.ts
rverdile: Yes, but I think 3 is probably unneeded so I took two of them out

dominikvagner (commented):
few improvement suggestions and questions 💭 
overall it looks great, good job! 👏🏼✨

looked a bit into the CI failure and I think it might be because of misconfigured environment vars in codebuild and in the action, fixed them in codebuild and made a [PR](https://github.com/content-services/content-sources-frontend/pull/559) that adjusts the action, after that gets merged and this rebased, hopefully it should go away 🙏🏼 



_edit: the PR was merged, so try rebasing and rerunning the nightly CI again_ ♻️

_playwright-tests/Integration/helpers/rhsmClient.ts
```diff
@@ -4,7 +4,7 @@ import { ExecReturn, killContainer, runCommand, startNewContainer } from './cont
 /**
  * Supported Operating System versions
  */
-export type OSVersion = 'rhel9' | 'rhel8';
+export type OSVersion = 'rhel9' | 'rhel8' | 'rhel9dev';
```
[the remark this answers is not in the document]
rverdile: it's from this list of container images: https://github.com/content-services/content-sources-frontend/blob/e6b8a8ea3a4b86ec9903fa92237899c62e61bc53/_playwright-tests/Integration/helpers/rhsmClient.ts#L12-L16

rverdile:
thanks all, updated

_playwright-tests/Integration/CanUpdateSystemWithTemplate.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+      expect(yumInstall?.exitCode).toBe(0);
+
+      // Now vim-enhanced should be installed
+      const exist = await regClient.Exec(['rpm', '-q', 'vim-enhanced']);
+      expect(exist?.exitCode).toBe(0);
+    });
+
+    await test.step('Remove system', async () => {
+      await regClient.Destroy(true);
+    });
```
dominikvagner, line 152: small comment to this: if the test fails somewhere in the previous steps, it might not reach this one and the client might not get destroyed 🙅🏼 _(not sure if the library the container helpers are built on, cleans up after itself by default or at all)_
this wouldn't probably be an issue in CI since after the runner gets to the end everything would be killed, but locally this could be annoying 😅 

this could be resolved by deleting this step and using the cleanup fixture at the begging, in a similar way as template cleanup is done, i.e.: 
```ts
cleanup.add(() => regClient.Destroy(true));
```
what this will do is add the `regClient.Destroy(true)` to the cleanup functions map, where everything in there will be run after the tests completes (either fails or passes) and that should prevent the hanging abandoned containers 🧹 
you can think of the cleanup add as of defer in Go 🧠😁

_playwright-tests/Integration/CanUpdateSystemWithTemplate.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+import { closePopupsIfExist, getRowByNameOrUrl } from '../UI/helpers/helpers';
+import { randomName } from '../UI/helpers/repoHelpers';
+
+const templateNamePrefix = 'integration_test_template';
+const templateName = `${templateNamePrefix}-${randomName()}`;
+const regClient = new RHSMClient('RHSMClientTest');
+let firstVersion;
+
+test.describe('Test System With Template', async () => {
+  test('Test System With Template', async ({ page, client, cleanup }) => {
```
dominikvagner, line 13: tiny insignificant nitpick: it would be nicer (look better in the report) if the name of the test and describe were different 💅🏼✨

dominikvagner (changes requested):
thanks for working on this and rebasing few times to take in my fixes made in parallel to this 🫶😇 
added two small nitpicks and one comment about an issue with the container cleanup 💭

_playwright-tests/Integration/CanUpdateSystemWithTemplate.spec.ts
swadeley: Hi @rverdile 
Downgrading a package is one way to create the need for an update. In my Satellite days, for an errata enumeration test,  I installed a package from CentOS to trigger the need for an update.

rverdile: That makes sense. In this case we would want changing the template date to make a new package version available (and cause the need for an update), so I'm not sure it would verify that, unless I'm misunderstanding.

dominikvagner (approved):
thanks for all the fixes! this looks awesome, nice job! 👏🏼😇
approved! ✅🚀

_to the updating of an installed package: I think we can do that in another test (or alter this one) once we figure out a stable way to do so. I am for merging this as it is right now 😅_

marusak (approved):
Thank you!

PR #566, opened 2025-07-14
HMS-8732, HMS-8733: UX review styling updates

Starle21:
## Summary

Updating UI based on UX review document:
- spacing issues impacting semantic grouping in Repositories and Templates pages
- following task to review: HMS-8832

### DIFF
--- a/src/Pages/Repositories/ContentListTable/ContentListTable.tsx
+++ b/src/Pages/Repositories/ContentListTable/ContentListTable.tsx
@@ -54,6 +54,7 @@ import { Outlet, useNavigate, useOutletContext, useSearchParams } from 'react-ro
 import useArchVersion from 'Hooks/useArchVersion';
 import { ADD_ROUTE, DELETE_ROUTE, EDIT_ROUTE, UPLOAD_ROUTE } from 'Routes/constants';
 import UploadRepositoryLabel from 'components/UploadRepositoryLabel/UploadRepositoryLabel';
+import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 
 const useStyles = createUseStyles({
   mainContainer: {
@@ -488,7 +489,7 @@ const ContentListTable = () => {
       >
         <Flex
           justifyContent={{ default: 'justifyContentSpaceBetween' }}
-          className='pf-v6-u-py-md pf-v6-u-px-lg'
+          className={`${spacing.pyMd} ${spacing.pxLg}`}
         >
           <ContentListFilters
             contentOrigin={contentOrigin}
@@ -553,7 +554,7 @@ const ContentListTable = () => {
               </Grid>
             </Hide>
             <Hide hide={countIsZero || isLoading}>
-              <Stack className='pf-v6-u-px-lg'>
+              <Stack className={spacing.pxLg}>
                 <Table
                   aria-label='Custom repositories table'
                   ouiaId='custom_repositories_table'
--- a/src/Pages/Repositories/ContentListTable/components/DeleteContentModal/DeleteContentModal.tsx
+++ b/src/Pages/Repositories/ContentListTable/components/DeleteContentModal/DeleteContentModal.tsx
@@ -33,6 +33,7 @@ import { isEmpty } from 'lodash';
 import useDeepCompareEffect from 'Hooks/useDeepCompareEffect';
 import { usePopularListOutletContext } from 'Pages/Repositories/PopularRepositoriesTable/PopularRepositoriesTable';
 import { ActionButtons } from 'components/ActionButtons/ActionButtons';
+import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 
 const useStyles = createUseStyles({
   removeButton: {
@@ -164,7 +165,7 @@ export default function DeleteContentModal() {
               Deleting these repositories will delete that content from their associated templates.
             </Alert>
           </Hide>
-          <div className='pf-v6-u-px-lg'>
+          <div className={spacing.pxLg}>
             <Content component='p'>Are you sure you want to delete these repositories?</Content>
           </div>
         </Stack>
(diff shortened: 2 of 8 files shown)


_playwright-tests/test-utils
```diff
@@ -1 +1 @@
-Subproject commit e3cc6178091e7644df4645e2a8e6515d59c21127
+Subproject commit e13ef79aca1d5f507e52cefbdc226f36e8cc10a5
```
dominikvagner, line 1: this is a downgrade, we definitely don't want to do that 🚫

build-tools
```diff
@@ -1 +1 @@
-Subproject commit 07aeb0dd02b76bd1c9c309b8ec0e96893e0ebe49
+Subproject commit 75adad05c9e22ff84c7d3b43564554a26f55a8a9
```
dominikvagner, line 1: ditto ⬆️

yarn.lock
```diff
@@ -4,22 +4,22 @@
 
 "@actions/core@^1.11.1":
   version "1.11.1"
-  resolved "https://registry.yarnpkg.com/@actions/core/-/core-1.11.1.tgz#ae683aac5112438021588030efb53b1adb86f172"
+  resolved "https://registry.npmjs.org/@actions/core/-/core-1.11.1.tgz"
```
dominikvagner, line 7: why was this done? the updates and the registry change 🚫  I don't think we want to do this :/

same goes for the `package-lock.json` we definitely don't want that as we are using yarn for package management 🙅🏼 

if you don't want to use yarn, then please add appropriate global git ignores and use that on your own without this being committed

_playwright-tests/test-utils
Starle21: ack

build-tools
Starle21: ack

yarn.lock
Starle21: ack

src/Pages/Repositories/ContentListTable/ContentListTable.tsx
```diff
@@ -556,7 +555,7 @@ const ContentListTable = () => {
               </Grid>
             </Hide>
             <Hide hide={countIsZero || isLoading}>
-              <>
+              <Stack className='pf-v6-u-px-lg'>
```
dominikvagner, line 558: the addition of a stack here is the culprit of the errors in the playwright tests 🔍 
due to how the stack behaves in here, it can break the page by overlapping the pagination button above it and makes it unclickable 🙅🏼 

2025-07-21-170640_hyprshot

two possible solutions come to mind:
1. use something else then Stack. maybe a plain div with the padding class, flex, or maybe a grid item since this is inside of a grid 💭 
2. change the wrapping Grid to a Stack and put the stuff above the table and the table itself into StackItems inside of it (tried this quickly and it seemed to work, but it is a bigger change) 😅 

so please try experimenting and fix this, as it is indeed broken in this way 🐛

dominikvagner: and also have a look at the templates page and other pages using tables, since this PR makes similar tweaks across multiple places 😅

dominikvagner (changes requested):
left couple of comments 😅  changes needed as this breakes the repositories page 🐛

src/components/ActionButtons/ActionButtons.tsx
```diff
@@ -0,0 +1,30 @@
+import { Button, Flex } from '@patternfly/react-core';
+import { FC } from 'react';
+
+type ActionButtonsProps = {
+  isAction: boolean;
+  onSave: () => void;
+  onClose: () => void;
+};
+
+export const ActionButtons: FC<ActionButtonsProps> = (props) => {
```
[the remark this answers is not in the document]
Starle21: ack

src/Pages/Repositories/ContentListTable/ContentListTable.tsx
```diff
@@ -486,11 +482,14 @@ const ContentListTable = () => {
...
       </Hide>
-      <Grid
+      <div
         data-ouia-component-id='content_list_page'
         className={countIsZero ? classes.mainContainer100Height : classes.mainContainer}
       >
-        <Flex className={classes.topContainer}>
+        <Flex
+          justifyContent={{ default: 'justifyContentSpaceBetween' }}
+          className='pf-v6-u-py-md pf-v6-u-px-lg'
```
xbhouse, line 491: these utility classes are used in a few places and it's not easy to interpret what they're for :confused: what do you think about adding a named style (ie: verticalMediumPadding or something) in useStyles and referencing them from there?

xbhouse:
could you also change the question icons in the AddContent modal to be without fill? for example, replacing the `<FormGroupLabelHelp aria-label='Add GPG Key help' />` with something like `<Button icon={<OutlinedQuestionCircleIcon />} variant='plain' aria-label='Add GPG Key help' />`

src/Pages/Repositories/ContentListTable/components/DeleteContentModal/DeleteContentModal.tsx
```diff
@@ -7,9 +7,8 @@ import {
   List,
```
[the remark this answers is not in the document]
Starle21: fixed.

src/Pages/Repositories/ContentListTable/ContentListTable.tsx
Starle21: I thought about it. 
In my first version, I even defined a utility file that contained all these patternfly defined classnames and variables, so we could reference it from one place and use it consistently.  And if patternfly decided to change it later, we would just change it in one place. 

Matej however didn't like it, he prefers for us to get rid of custom defined useStyles altogether. 

However I can add it back in, as I think it would be cleaner as well. Plus I can make it such we don't need to use useStyles, so it would pass :)

dominikvagner (approved):
looks great! ✨ and tests are now green 🚀 
thanks! 

approved! ✅

PR #568, opened 2025-07-14
RHINENG-18926: Add patch tests and helper

niyazRedhat:
## Summary

1. Patch test for exporting System.
2. Patch test Plan remediation button can not be clicked when no system is selected 

## Testing steps
1. Navigate to the Systems all page.
2. Export by CSV/JSON.
expected_results:
2. Exported data will be downloaded.

Remediation button test
1. Navigate to the Systems all page.
2. plan remediation button can not be clicked

### DIFF
--- a/_playwright-tests/UI/ExportSystem.spec.ts
+++ b/_playwright-tests/UI/ExportSystem.spec.ts
@@ -1,0 +1,33 @@
+import { test, expect } from 'test-utils';
+import { navigateToSystems } from './helpers/navHelpersPatch';
+
+test.describe('Export Systems List', () => {
+  test('Navigate to systems, export list to CSV and JSON', async ({ page }) => {
+    await navigateToSystems(page);
+    await test.step('export list to CSV', async () => {
+      await page.getByRole('button', { name: 'Export' }).click();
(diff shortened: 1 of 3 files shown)


swadeley:
ok-to-test

xbhouse:
you can run `yarn format` to run prettier locally, in this case the small format fixes will be made automatically once you run that :smiley:

_playwright-tests/UI/ExportSystem.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { test, expect } from 'test-utils';
+import { navigateToSystems } from './helpers/navHelpersPatch';
+
+test('EXPORT', async ({ page }) => {
+  await navigateToSystems(page);
+
+  await page.getByRole('button', { name: 'Export' }).click();
+  await page.getByRole('menuitem', { name: 'Export to CSV' }).click();
+
+  const preparingAlert = page.locator('div.pf-v5-c-alert.pf-m-info.notification-item');
```
xbhouse, line 10: we try to only use locators that may change (like ones with patternfly in the name) as a last resort. is there another way to target the notification by using the text maybe?

swadeley (commented):
Hi, use test-step rather than inline comments

PR #578, opened 2025-07-16
Test: stop prettier checking style guide

swadeley:
## Summary

stop prettier checking style guide

because it logs "warn" in PR test runs.

## Testing steps

tests pass

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,2 @@
 _playwright-tests/test-utils/**/*
+_playwright-tests/style_guide.md


swadeley:
> Can you show such output? Wouldn't it be actually beneficial if it warns you?

You can see it here:
https://github.com/content-services/content-sources-frontend/actions/runs/16318225159/job/46089182407?pr=577

xbhouse:
ran `yarn format` on the style guide and looks like it's just removing some newlines. it could be good to keep the guide in the format check, maybe we should check in those changes to the guide instead?

swadeley:
> ran `yarn format` on the style guide and looks like it's just removing some newlines. it could be good to keep the guide in the format check, maybe we should check in those changes to the guide instead?

OK, closing, thank you

PR #580, opened 2025-07-17
Feat: tag filtering in add system modal

darkeriossss:
## Summary
Add tag filtering to Add systems to template modal

## Testing steps
1. Navigate to Template detail
2. Click on the Systems tab
3. Open Add systems modal
4. Try out filtering by tags

### DIFF
--- a/src/Pages/Templates/TemplateDetails/components/AddSystems/AddSystemModal.tsx
+++ b/src/Pages/Templates/TemplateDetails/components/AddSystems/AddSystemModal.tsx
@@ -2,6 +2,9 @@ import {
   AlertVariant,
   Bullseye,
   Button,
+  Dropdown,
+  DropdownItem,
+  DropdownList,
   Flex,
   FlexItem,
   Grid,
@@ -11,6 +14,7 @@ import {
   ModalFooter,
   ModalHeader,
   ModalVariant,
+  MenuToggle,
   Pagination,
   PaginationVariant,
   Spinner,
@@ -22,7 +26,7 @@ import { InnerScrollContainer } from '@patternfly/react-table';
 import { useEffect, useMemo, useState } from 'react';
 import { createUseStyles } from 'react-jss';
 import Hide from 'components/Hide/Hide';
-import { SearchIcon, SyncAltIcon } from '@patternfly/react-icons';
+import { FilterIcon, SearchIcon, SyncAltIcon } from '@patternfly/react-icons';
 import useDebounce from 'Hooks/useDebounce';
 import { useNavigate, useParams } from 'react-router-dom';
 import useRootPath from 'Hooks/useRootPath';
@@ -45,6 +49,7 @@ import {
 import ModalSystemsTable from './ModalSystemsTable';
 import ConditionalTooltip from 'components/ConditionalTooltip/ConditionalTooltip';
 import useNotification from 'Hooks/useNotification';
+import TagsFilter from 'components/TagsFilter/TagsFilter';
 
 const useStyles = createUseStyles({
   mainContainer: {
@@ -70,9 +75,14 @@ const useStyles = createUseStyles({
     paddingTop: '0',
     paddingBottom: '0',
   },
+  fullWidth: {
+    width: 'auto',
+    maxWidth: 'unset',
+  },
 });
 
 const perPageKey = 'templatesPerPage';
+type FilterType = 'Name' | 'Tags';
 
 export default function AddSystemModal() {
   const queryClient = useQueryClient();
@@ -90,6 +100,9 @@ export default function AddSystemModal() {
   const { notify } = useNotification();
   const [pollCount, setPollCount] = useState(0);
   const [polling, setPolling] = useState(false);
+  const [filterType, setFilterType] = useState<FilterType>('Name');
+  const [selectedTags, setSelectedTags] = useState<string[]>([]);
+  const [filterOpen, setFilterOpen] = useState(false);
 
   useEffect(() => {
     if (!selectedList.size) {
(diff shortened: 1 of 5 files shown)


src/Pages/Templates/TemplateDetails/components/AddSystems/AddSystemModal.tsx
```diff
@@ -235,6 +249,32 @@ export default function AddSystemModal() {
...
+          <TextInput
+            id='search'
+            type='search'
+            className={classes.fullWidth}
+            customIcon={<SearchIcon />}
+            ouiaId='filter_search'
+            placeholder='Filter by name'
+            value={searchQuery}
+            onChange={(_event, value) => {
+              console.log('value', value);
```
jlsherrill, line 267: can remove this :)

src/components/TagsFilter/TagsFilter.tsx
```diff
@@ -0,0 +1,204 @@
...
+      item: {
+        meta: {
+          tag: { key, value },
+        },
+      },
+    };
+  });
+
+  console.log('tags', tagsData, isLoading);
+  console.log('grouped', groupedByNamespace);
```
jlsherrill, line 79: these too

jlsherrill:
One small issue i see:

1.  have more than 10 tags
2. select 1 tag:
image
3.  click "10 more tags available"
4. Notice it doesn't indicate any tags selected in the upper left:
image
5.  select a 2nd tag, notice it says '1 selected' even though 2 are:
image

_playwright-tests/UI/TemplateCRUD.spec.ts
```diff
@@ -85,6 +85,125 @@ test.describe('Templates CRUD', () => {
       await page.getByRole('button', { name: 'Next', exact: true }).click();
       await page.getByRole('button', { name: 'Confirm changes', exact: true }).click();
     });
+    await test.step('Filter systems in Add systems modal', async () => {
+      await page.route('**/api/patch/v3/templates/*/systems?offset=0&limit=20', async (route) => {
```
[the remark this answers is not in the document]
dominikvagner: agree, this would be better suited as an integration test where a client would actually register to insights and this would be verified in a real scenario 👍🏼

_playwright-tests/UI/TemplateCRUD.spec.ts
```diff
@@ -85,6 +85,125 @@ test.describe('Templates CRUD', () => {
...
+        await route.fulfill({ response, status: 200, json });
+      });
+      const rowTemplate = await getRowByNameOrUrl(page, `${templateName}-edited`);
+      await rowTemplate.getByRole('button', { name: templateName }).click();
+      await page.getByTestId('systems_tab').click();
+      await page.getByRole('button', { name: 'Assign to systems' }).click();
+      await expect(page.getByRole('row', { name: 'test_host' }).first()).toBeVisible();
+      await page.getByTestId('OUIA-Generated-MenuToggle-1').click();
+      await page.getByTestId('filter_Tags').locator('button').click();
+      await page.getByTestId('system_modal').getByTestId('OUIA-Generated-MenuToggle-2').click();
```
dominikvagner, line 197: are these (and other elements in this test) really not targetable by something else than test IDs? we prefer not to use them unless absolutely necessary, check out what other selectors are there in playwright and try some of those 👍🏼 
sidenote: almost anything can be targeted with `getByRole` or an appropriate chain of good selectors 😅

_playwright-tests/UI/TemplateCRUD.spec.ts
```diff
@@ -85,6 +85,125 @@ test.describe('Templates CRUD', () => {
...
+      const tagSelector = page.locator('label').filter({ hasText: 'abc=xyz' });
+      await expect(tagSelector).toBeVisible();
+      const systemsRequestPromise = page.waitForRequest('**/api/patch/v3/systems**');
+      await tagSelector.click();
+      await systemsRequestPromise;
```
dominikvagner, line 202: the systems request promise is only used one, no need to define it as a variable 🧹 

and also we found that the pattern of clicking a button and then awaiting the promise afterwards can sometimes be flaky, it's better to run these concurrently with `Promise.All` [[example](https://github.com/content-services/content-sources-frontend/blob/55676ed847e00d3e122798fbcf3517ccffb22277/_playwright-tests/UI/SnapshotRepo.spec.ts#L44-L51)] ❄️😅

src/Pages/Templates/TemplateDetails/components/AddSystems/AddSystemModal.tsx
```diff
@@ -67,9 +72,14 @@ const useStyles = createUseStyles({
     paddingTop: '0',
     paddingBottom: '0',
   },
+  fullWidth: {
+    width: 'auto',
+    maxWidth: 'unset',
+  },
```
dominikvagner, line 78: there is still some custom CSS in our app, but the goal going forward is to minimize that to or close to zero 🗑️ 
is there no way to achieve something similar (so the filter text doesn't get cut off) as this with some layout components or built-in classes in PF6? could you please look at that 🙏🏼😅

dominikvagner (commented):
added a couple of comments/suggestions 💭 
and also the lint issue in this PR will need to be fixed, this one should be autofixable by running `yarn lint:fix` 😅

src/Pages/Templates/TemplateDetails/components/AddSystems/AddSystemModal.tsx
xmicha82: sadly i wasn't able to get it working using pf classes

dominikvagner (dismissed):

dominikvagner (approved):
looks good! ✨ 
great job, thanks! 👏🏼😇

approved ✅

PR #585, opened 2025-07-22
Test: Do not check for bulk delete URL when deleting a single repo

swadeley:
## Summary

This step is deleting a single repo not doing a bulk delete.

The URL will look like this for a single repo if you access the delete modal using the kebab menu:
`content/repositories/delete?repoUUID=2582bfac-7319-4a6e-bf13-34f6e01aab28`

However if you select the row and then use the dropdrown menu, the URL will just be:
`insights/content/repositories/delete`

## Testing steps

tests pass

### DIFF
--- a/_playwright-tests/UI/SnapshotRepo.spec.ts
+++ b/_playwright-tests/UI/SnapshotRepo.spec.ts
@@ -94,8 +94,7 @@ test.describe('Snapshot Repositories', () => {
 
       await Promise.all([
         page.waitForResponse(
-          (resp) =>
-            resp.url().includes('bulk_delete') && resp.status() >= 200 && resp.status() < 300,
+          (resp) => resp.url().includes('delete') && resp.status() >= 200 && resp.status() < 300,
         ),
         page.getByRole('button', { name: 'Remove' }).click(),
       ]);


_playwright-tests/UI/SnapshotRepo.spec.ts
```diff
@@ -91,15 +91,7 @@ test.describe('Snapshot Repositories', () => {
...
       await expect(page.getByText('Remove repositories?')).toBeVisible();
-
-      await Promise.all([
-        page.waitForResponse(
-          (resp) =>
-            resp.url().includes('bulk_delete') && resp.status() >= 200 && resp.status() < 300,
-        ),
-        page.getByRole('button', { name: 'Remove' }).click(),
-      ]);
-
```
dominikvagner, line 102: I think this would still be nice to have as without this, it won't assert/verify the response at all 🤔 

> This step is deleting a single repo not doing a bulk delete.
⬆️ since this is the case for this step, why not just check for a `delete` instead of a `bulk_delete` in the URL?

dominikvagner (commented):
one comment regarding the removal of the response verification 💭

swadeley:
> one comment regarding the removal of the response verification 💭
> 
> _also would be nice to make PRs not from branches based on other open PRs 😉😁_

Sorry, I'll rebase

_playwright-tests/UI/SnapshotRepo.spec.ts
swadeley: I thought Playwright takes care of waiting for a button to be enabled?

dominikvagner (approved):
looking good now! ✨ 
thanks! 😇 

approved! ✅

PR #592, opened 2025-07-23
HMS-5260: fix text style issues and add question mark with popover 

Dugowitch:
## Summary
- add missing punctuation
- rename button to 'Create template' (was 'Add content template')
- replace 'Version(s)' with 'OS version(s)' (not only 'OS' as suggested to stay consistent with [HMS-5263](https://issues.redhat.com/browse/HMS-5263))
- add question mark button with popover to the header of Templates


jlsherrill:
https://issues.redhat.com/browse/HMS-5260

Dugowitch:
The integration test failure is expected. It fails because button label was changed and it cannot find the element, should pass when stage is redeployed to this commit

Dugowitch:
hey @xbhouse, i addressed the test. could you re-review please

PR #593, opened 2025-07-23
Test: add a gate for RBAC tests

dominikvagner:
## Summary
This adds a new environment variable that functions as a gate for running RBAC tests. With this they are going to be opt-in now, which will be handy for running tests locally as it will cut down the time for setup and authentication.
This will also help newcomers and people that want to try playwright through our repo, as they won't need all of the other user's credentials.
The RBAC tests and the full setup will still be run in CI.

## Testing steps
- RBAC tests should be run and pass in CI.
- RBAC tests shouln't run locally unless the RBAC flag (environment variable) is set to true.


xbhouse (approved):
looks good!

PR #597, opened 2025-07-29
Add NodeJS Lifecycle Dates test

swadeley:
## Summary

Test that RHEL9 nodejs modules have lifecycle dates

## Testing steps

test passes


jlsherrill:
https://issues.redhat.com/browse/HMS-5822

_playwright-tests/Integration/NodeJSLifecycleDates.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+
+      expect(packageTypeStream?.startDate).toBeDefined();
+      expect(packageTypeStream?.endDate).toBeDefined();
+      expect(moduleTypeStream?.startDate).toBeDefined();
+      expect(moduleTypeStream?.endDate).toBeDefined();
+
+      expect(new Date(packageTypeStream?.startDate || '').getTime()).not.toBeNaN();
+      expect(new Date(packageTypeStream?.endDate || '').getTime()).not.toBeNaN();
+      expect(new Date(moduleTypeStream?.startDate || '').getTime()).not.toBeNaN();
+      expect(new Date(moduleTypeStream?.endDate || '').getTime()).not.toBeNaN();
```
marusak, line 53: How often do these change? Can we predict/manage these values? This test makes is very hand-wavy and pretty much any mistake would not be caught (e.g. in Patch right now there is a bug that shows 1.1.1970 for updates, yet this test would be happy with that).

swadeley: Hi @marusak , I added an assert for a minimum start date. Waiting for @jlsherrill's  review now.

_playwright-tests/Integration/NodeJSLifecycleDates.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+
+      expect(repositoryModuleStreams.length).toBeGreaterThan(0);
+
+      const repositoryNodejsModule = repositoryModuleStreams.find((mod) =>
+        mod.moduleName?.includes('nodejs'),
+      );
+      expect(repositoryNodejsModule).toBeDefined();
+
+      expect(repositoryNodejsModule?.streams).toBeDefined();
+      expect(repositoryNodejsModule?.streams?.length).toBeGreaterThan(0);
```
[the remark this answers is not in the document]
swadeley: I will discuss with Justin

_playwright-tests/Integration/NodeJSLifecycleDates.spec.ts
marusak: As discussed during the standup, why not using exact dates? They won't change during lifetime of RHEL release and they would provide much stronger tests.

jlsherrill: there probably is some value in checking for a specific date, although IDK that our api would ever return a date improperly, and this wouldn't stop users of this api (IB) to display the date improperly if it was missing.  

These dates shouldn't change.  For the module stream one, you'd probably want to choose a specific version:
```
      const moduleTypeStream = packageSources.find((source) => source.type === 'module' && source.version = "19.0" );
```
as an example, not tested.

_playwright-tests/Integration/NodeJSLifecycleDates.spec.ts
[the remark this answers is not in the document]
jlsherrill: I think this is really meant as an integration test for the roadmap api and this stream info isn't really coming from that and should be tested elsewhere (such as  backend unit tests).  So really i think we could drop this test altogether since its not checking lifecycle dates.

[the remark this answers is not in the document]
swadeley: @jlsherrill I have removed the nodejs module lifecycle date test.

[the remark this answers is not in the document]
swadeley: 'Check NodeJS Date Format' ?

_playwright-tests/Integration/NodeJSLifecycleDates.spec.ts
swadeley: OK, added that, thank you

jlsherrill: thats oaky, they can have different dates

_playwright-tests/Integration/NodeJSLifecycleDates.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+      expect(new Date(packageTypeStream?.endDate || '').getTime()).not.toBeNaN();
+      expect(new Date(moduleTypeStream?.startDate || '').getTime()).not.toBeNaN();
+      expect(new Date(moduleTypeStream?.endDate || '').getTime()).not.toBeNaN();
+
+      const expectedMinDate = new Date('2022-05-17').getTime();
+      expect(new Date(packageTypeStream?.startDate || '').getTime()).toBeGreaterThanOrEqual(
+        expectedMinDate,
+      );
+      expect(new Date(moduleTypeStream?.startDate || '').getTime()).toBeGreaterThanOrEqual(
+        expectedMinDate,
```
jlsherrill, line 61: we may want to check that its the exact date as what is expected, although really its unlikely to matter.  If the integration breaks it'd likely be blank, if something else goes really wrong it might be 1970.  If the date were to change in the roadmap api but still be greater than 2022, we'd just be propagating the date they are giving us so honestly that'd be okay.   I'll leave it up to you.  ACK :)

jlsherrill (approved):

_playwright-tests/Integration/NodeJSLifecycleDates.spec.ts
swadeley: Hi, lets make it strict and see how often it fails. At least we would know around the same time as Image Builder team.

_playwright-tests/Integration/NodeJSLifecycleDates.spec.ts
```diff
@@ -0,0 +1,59 @@
+import {
+  test,
+  expect,
+  RpmsApi,
+  ApiSearchRpmResponse,
+} from '../test-utils/_playwright-tests/test-utils/src';
```
dominikvagner, line 6: this import would be better if it used the custom global path to test utils, which is just `test-utils`, rather than the relative path to the source folder for it 🧹

dominikvagner (commented):
@marusak this test does depend on a stage backend where the appstream repo is snapshotted _(since that's a big one with 21.5k packages, we don't want to import it in the local CI)_, so I think it's okay to have it here even if it's a pure API test 😅 

one comment regarding the import path, other than that this looks good to me 👍🏼🚀
nice job!

PR #598, opened 2025-07-29
HMS-8773: Fix filter on repo status

swadeley:
## Summary

[HMS-8773] Filter on repo status: Change "Pending" to "In progress"

## Testing steps

tests pass and one screen capture

### DIFF
--- a/src/Pages/Repositories/ContentListTable/components/ContentListFilters.tsx
+++ b/src/Pages/Repositories/ContentListTable/components/ContentListFilters.tsx
@@ -64,7 +64,7 @@ const useStyles = createUseStyles({
   },
 });
 
-const statusValues = ['Invalid', 'Pending', 'Unavailable', 'Valid'];
+const statusValues = ['Invalid', 'In progress', 'Unavailable', 'Valid'];
 export type Filters = 'Name/URL' | 'Version' | 'Architecture' | 'Status';
 
 const ContentListFilters = ({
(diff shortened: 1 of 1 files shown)


swadeley:
image

src/Pages/Repositories/ContentListTable/components/ContentListFilters.tsx
```diff
@@ -64,7 +64,7 @@ const useStyles = createUseStyles({
   },
 });
 
-const statusValues = ['Invalid', 'Pending', 'Unavailable', 'Valid'];
+const statusValues = ['Invalid', 'In progress', 'Unavailable', 'Valid'];
```
xbhouse, line 67: since the API only recognizes `pending` and not `in progress` as a status, we'd also need to map that here, something like this:

```
const statusToApiMap: Record<string, string> = {
  'In progress': 'Pending',
};
```

then in the useEffect that calls setFilterData, you can map to the correct status in the api filters:

```
useEffect(() => {
    setFilterData({
      searchQuery: debouncedSearchQuery,
      versions: getLabels('version', debouncedSelectedVersions),
      arches: getLabels('arch', debouncedSelectedArches),
      statuses: debouncedSelectedStatuses.map((status) => statusToApiMap[status] || status),
    });
  }, [
    debouncedSearchQuery,
    debouncedSelectedVersions,
    debouncedSelectedArches,
    debouncedSelectedStatuses,
  ]);
```

PR #599, opened 2025-07-30
Update Add "Update" to title word list

swadeley:
## Summary

As Konflux opens PRs with title starting "Update" I think we should add it to the allowed list.

## Testing steps

peer review & title check test passes

### DIFF
--- a/.github/workflows/title.yml
+++ b/.github/workflows/title.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
       - uses: deepakputhraya/action-pr-title@master
         with:
-          regex: '^Build|Test|^(\w+-\d+(,\w+-\d+)*): .+' # Regex the title should match.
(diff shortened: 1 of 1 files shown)


swadeley:
> would this work if the PR title started with `Update ...` without a colon? 🤔 I think that Konflux doesn't put the colon after the update and the task would still fail on those, am I wrong? 💭

swadeley:
> > would this work if the PR title started with `Update ...` without a colon? 🤔 I think that Konflux doesn't put the colon after the update and the task would still fail on those, am I wrong? 💭
> 
> I will test that now

PR Validation / Title (Successful in 3s)

dominikvagner (approved):
ack! thx! 👍🏼🚀

swadeley:
Closing in favour of:

 Build: Add Build to red-hat-konflux PR title #620

PR #603, opened 2025-07-30
Test: configure proxy for rhc in stage

rverdile:
## Summary
The proxy was not correctly being appended to the insights-client config when registering rhc in stage


xbhouse (approved):
catted that file before and after your changes, proxy is there now :) ack!

PR #610, opened 2025-08-11
HMS-8961: QE migrate test_subscription_check

swadeley:
## Summary

[ [HMS-8961] QE: migrate test_subscription_check](
https://issues.redhat.com/browse/HMS-8961)

## Testing steps

tests pass


jlsherrill:
https://issues.redhat.com/browse/HMS-8961

dominikvagner:
> i tried logging in to my locally running frontend with the no-subs user and do see the `Create template` button enabled. i don't see it enabled in stage though with that user. i wonder could this have something to do with using a local instance of candlepin with the devel org? @jlsherrill

yeah I think this won't work with the locally deployed backend, the `devel_org` setting in the config makes the `subscription_check/` endpoint return true for all users [[1](https://github.com/content-services/content-sources-backend/pull/794)] 🔍 

we could make this test an integration one, so that it would go directly against stage and than this should work as expected 😅

swadeley:
Hi, integration tests passed.

_playwright-tests/auth.setup.ts
```diff
@@ -107,6 +108,25 @@ setup.describe('Setup Authentication States', async () => {
     await logout(page);
   });
 
+  setup('Authenticate no-subs user and save state', async ({ page }) => {
+    setup.skip(!process.env.RBAC, `Skipping as the RBAC environment variable isn't set to true.`);
```
xbhouse, line 112: we don't need to authenticate this user unless we're running integration tests, so this auth step should be skipped if either `RBAC` or `INTEGRATION` is set to true, similar to the skip statement you have in the NoSubsUserTest

dominikvagner: \* _if either `RBAC` or `INTEGRATION` is set to **false**_

.github/workflows/playwright-actions.yml
```diff
@@ -58,6 +58,16 @@ jobs:
             exit 1
           fi
 
+          if [ -z "$NO_SUBS_USER_USERNAME" ]; then
```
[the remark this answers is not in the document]
swadeley: OK, will do

_playwright-tests/auth.setup.ts
swadeley: OK, changes made, thank you for the reviews.

swadeley:
rebased

PR #613, opened 2025-08-13
Test: Fix removing a template warning message assert

swadeley:
## Summary

    Message is not displayed instantly

so need to use    "await expect" with "toBeVisible()"

## Testing steps
tests pass


xbhouse (approved):
ack!

PR #615, opened 2025-08-14
Test: fix oom problem rhsmclient

swadeley:
## Summary

Simplify dnf commands in Integration/CanUpdateSystemWithTemplate.spec.ts to reduce memory use causing OOM failure in container.

## Testing steps

'Test System With Template' test passes


swadeley:
I plan to close this PR in favour of:

 [Test: Fix container name with random suffix #619](https://github.com/content-services/content-sources-frontend/pull/619) 

I can make another PR to simplify the dnf commands if required.
Product UUID problem is out of scope for this issue and I will get back to that.

PR #622, opened 2025-08-21
HMS-8883: spinner behavior triggered only on data fetching

Starle21:
## Summary
Changing spinner's behavior:

- Spinner's behavior placed in the top right corner of repositories table was triggered on 4 occasions (isFetching, repositoryParamsLoading, isIntrospecting, isDeletingItems). 
- Plus even when there was no action, spinner animation kept running. 
- Except for isFetching, the other actions have different UX feedback tied to them and triggering the spinner was redundant for them.

This is now changed that spinner appears only when:
-  there is content list query data fetching going on, 
-  or content list query is polling data. 

It appears mainly on a slow connection when a user navigates between filters of the table.

For the future it would be great to redefine the placement of the spinner and the whole UX associated with it. 

## Testing steps
Throttle your connection and navigate between filters of the table (custom, redhat repos).

### DIFF
--- a/src/Pages/Repositories/ContentListTable/ContentListTable.tsx
+++ b/src/Pages/Repositories/ContentListTable/ContentListTable.tsx
@@ -68,9 +68,6 @@ const useStyles = createUseStyles({
   bottomContainer: {
     justifyContent: 'space-between',
   },
-  invisible: {
-    opacity: 0,
-  },
   checkboxMinWidth: {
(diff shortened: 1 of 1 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-8883

xbhouse (approved):
lgtm! much less spinning :smile: :tada:

PR #627, opened 2025-08-26
HMS-9166

katarinazaprazna:

### DIFF
--- a/src/Pages/Repositories/ContentListTable/components/AddContent/AddContent.tsx
+++ b/src/Pages/Repositories/ContentListTable/components/AddContent/AddContent.tsx
@@ -1,24 +1,24 @@
 import {
+  Alert,
+  Button,
+  Dropdown,
+  DropdownItem,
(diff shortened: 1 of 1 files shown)


xbhouse:
> awesome, I see no more flickering! ✨ great work, thanks! 👏🏼🚀
> 
> just to be safe: @xbhouse I know you made this change because it was related to that modal bug in patternfly we ran into when migrating 🔍
> 
> ```diff
> - popperProps={versionOpen ? { appendTo: 'inline' } : undefined}
> + popperProps={{ appendTo: 'inline' }}
> ```
> 
> do you remember why it was appended to the modal only if it was open, tests seem to pass with these changes and the targeting seems to work as well, was there something else? 🤔

@dominikvagner i don't think there was a reason for appending only when open, since this change fixes the targeting also **and** the flickering, this looks good to me :)

xbhouse (approved):
ack from me!

PR #628, opened 2025-08-27
HMS-5829: support for templates with shared EPEL

xbhouse:
## Testing steps


src/Pages/Repositories/ContentListTable/ContentListTable.tsx
rverdile, line 1: One comment that isn't an issue with your PR specifically, so if you want we can open a new card for this. 

The UI gives me the option to delete the community EPEL snapshots (it doesn't actually work in the API though).

image

src/Pages/Templates/TemplatesTable/components/AddTemplate/AddTemplate.tsx
rverdile, line 1: The "Other repositories" steps says select "custom" repositories. I guess it should say "custom or EPEL"?

image

rverdile (commented):
just two small comments, everything else looks good

PR #630, opened 2025-08-29
HMS-8608: display delete template modal over template detail page

Starle21:
## Summary
Previously when on a template details page, clicking the delete action:
- redirected to the templates list page where it was overlayed with the delete template modal. 

This might have confused users, especially when cancelling the action. 

Now the delete modal:
- is showed over the template details
- and only after the delete confirmation a user is navigated to the templates list page.

And as expected, when the delete action is triggered from the templates list page, the delete modal is overlayed over the templates list.

## Testing steps
- navigate to a template detail, trigger delete action there, test out cancel and delete
- you can also test out that the modal is layered over the templates list page when it happens to be triggered from there

### DIFF
--- a/src/Pages/Templates/TemplatesTable/components/DeleteTemplateModal.test.tsx
+++ b/src/Pages/Templates/TemplatesTable/components/DeleteTemplateModal.test.tsx
@@ -4,6 +4,7 @@ import { ReactQueryTestWrapper, defaultSystemsListItem, defaultTemplateItem } fr
 import DeleteTemplateModal from './DeleteTemplateModal';
 import { useListSystemsByTemplateId } from 'services/Systems/SystemsQueries';
 import { useFetchTemplate } from 'services/Templates/TemplateQueries';
+import { DETAILS_ROUTE } from 'Routes/constants';
 
 jest.mock('react-query', () => ({
   ...jest.requireActual('react-query'),
@@ -15,6 +16,9 @@ jest.mock('react-router-dom', () => ({
     templateUUID: `${defaultTemplateItem.uuid}`,
   }),
   useNavigate: () => jest.fn,
+  useLocation: () => ({
+    pathname: `/mocked/${DETAILS_ROUTE}`,
+  }),
 }));
 
 jest.mock('Hooks/useRootPath', () => () => 'someUrl');
--- a/src/Routes/index.tsx
+++ b/src/Routes/index.tsx
@@ -143,9 +143,7 @@ export default function RepositoriesRoutes() {
         </Route>
         {!rbac?.templateRead ? (
           <Route path={TEMPLATES_ROUTE} element={<NoPermissionsPage />} />
-        ) : (
-          ''
-        )}
+        ) : null}
         <Route
           path={`${TEMPLATES_ROUTE}/:templateUUID/${DETAILS_ROUTE}`}
           element={<TemplateDetails />}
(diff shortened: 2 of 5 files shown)


swadeley:
Hi @Starle21 

The Delete confirmation modal does remain over the Template Details page, however I see popup error "Unable to find systems with the given id xxxx" even thoguh the template is not associated to a system. Might be unrelated to your changes.

src/Pages/Templates/TemplatesTable/components/DeleteTemplateModal.test.tsx
```diff
@@ -15,6 +15,9 @@ jest.mock('react-router-dom', () => ({
     templateUUID: `${defaultTemplateItem.uuid}`,
   }),
   useNavigate: () => jest.fn,
+  useLocation: () => ({
+    pathname: '/mocked/details',
```
katarinazaprazna, line 19: To make the test more robust, we could consider using pathname: `/mocked/${DETAILS_ROUTE}` instead

src/Routes/index.tsx
```diff
@@ -44,6 +44,7 @@ import UploadContent from 'Pages/Repositories/ContentListTable/components/Upload
 import DeleteSnapshotsModal from 'Pages/Repositories/ContentListTable/components/SnapshotListModal/DeleteSnapshotsModal/DeleteSnapshotsModal';
 import AdminFeaturesTable from 'Pages/Repositories/AdminFeaturesTable/AdminFeaturesTable';
 import PopularRepositoriesTable from 'Pages/Repositories/PopularRepositoriesTable/PopularRepositoriesTable';
+// import { DeleteTemplateModalTest } from 'Pages/Templates/TemplatesTable/components/DeleteTemplateTest';
```
katarinazaprazna, line 47: This comment was probably added by accident

katarinazaprazna:
I've left a few comments with some suggestions. Overall, this is a great PR that significantly improves the UX :)

katarinazaprazna:
Thanks for making those changes @Starle21. I'm also seeing the error @swadeley  mentioned when I run the project locally. It would be great if someone else could confirm this as well. Other than that, it's an ack from me :)

dominikvagner (changes requested):
the code looks good overall and the delete modal not jumping to the template list is great 👍🏼 _(my only comment would be to the changes in `src/Routes/index.ts` where some of them seem kinda unrelated/unnecessary)_

but I have 2 other comments 💭 
- the content of the template details page seems to change when the display is shown and then reloaded when it's closed, so if that could be fixed that would be nice 😅 
- the more important issue is that the edit and delete modals now behave differently, the edit still takes you to the template list and doesn't stay on the details page, would be great if the edit would work the same, i.e.: as the delete modal works after this change 😄 

try to change the edit modal in the same way and have a look if it's possible to prevent the content changing behind the modal 😅 


https://github.com/user-attachments/assets/aefb7012-12df-45b4-87e0-9d14db3e0d74

Starle21:
> * the content of the template details page seems to change when the display is shown and then reloaded when it's closed, so if that could be fixed that would be nice 😅

This is something that in the current routes setup cannot be easily changed. There are other subpaths that are being shown at the same time with the parent path, that includes the delete action. So to be able to show the modal over the parent path, the behavior is expected.


> * the more important issue is that the edit and delete modals now behave differently, the edit still takes you to the template list and doesn't stay on the details page, would be great if the edit would work the same, i.e.: as the delete modal works after this change 😄

This is the question of scope. Since I didn't look into the edit code, which may be different, I wanted to avoid scope creep. I am aware of the edit behavior not being consistent, and I filed [a ticket for that](https://issues.redhat.com/browse/HMS-9288).

katarinazaprazna (changes requested):
I noticed that when you delete a template from the details page, the **Actions** dropdown doesn't close after you click the Delete action. It stays open, displayed above the modal overlay. Could you please take a look into this? Thanks

Image

Starle21:
note: security/snyk manually marked as successful, since the AI was throwing false positive Open Redirect, even though the input was sanitized to make sure it is a valid UUID.

PR #631, opened 2025-08-29
HMS-5147: verify users can't access content they aren't allowed to

swadeley:
## Summary

Add No Cross Org Access test
    
        HMS-5147 verify users can`t access custom and red hat content they aren`t allowed to

## Testing steps

tests pass

### DIFF
--- a/_playwright-tests/Integration/NoCrossOrgAccess.spec.ts
+++ b/_playwright-tests/Integration/NoCrossOrgAccess.spec.ts
@@ -1,0 +1,229 @@
+import { test, expect, RepositoriesApi, SnapshotsApi } from 'test-utils';
+import { RHSMClient } from './helpers/rhsmClient';
+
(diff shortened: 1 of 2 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-5147

rverdile (commented):
I see this test verifies that a user with no subs cannot access content, and that the user with no subs can't access the content of a different org. Should it also verify that a user with subs cannot access the content of a different org?

swadeley:
> I see this test verifies that a user with no subs cannot access content, and that the user with no subs can't access the content of a different org. Should it also verify that a user with subs cannot access the content of a different org?

Hi @rverdile , I think that is covered by the IQE test `test_snapshot_listing_filter_by_org_id`  that is being migrated in https://issues.redhat.com/browse/HMS-8895

rverdile (approved):

PR #636, opened 2025-09-04
HMS-8895: Add cross-org snapshot access control validation

mayurilahane:
- Add integration test to validate that users from different organizations 
cannot access repository configuration files using snapshot UUIDs from 
other orgs.

- Test uses stable_sam_stage user to create repo and verifies 
403 Forbidden response when cross-org user attempts access.


jlsherrill:
https://issues.redhat.com/browse/HMS-8895

PR #637, opened 2025-09-04
Test: Add try block to subs-man refresh

swadeley:
## Summary

   subscription-manager refresh sometimes fails with error code 70.
   This could be caused by poor connectivity within the cluster.
   This change will mitigate that and provide logs.

## Testing steps

These two pass:
_playwright-tests/Integration/AssociatedTemplateCRUD.spec.ts
_playwright-tests/Integration/CanUpdateSystemWithTemplate.spec.ts

### DIFF
--- a/_playwright-tests/Integration/AssociatedTemplateCRUD.spec.ts
+++ b/_playwright-tests/Integration/AssociatedTemplateCRUD.spec.ts
@@ -4,7 +4,7 @@ import {
   cleanupTemplates,
   randomName,
 } from '../test-utils/_playwright-tests/test-utils/src';
-import { RHSMClient } from './helpers/rhsmClient';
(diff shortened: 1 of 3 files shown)


marusak:
Did it fail at least once (or at least locally)? To see what kind of output it provides.
Also, what about extracting it into a helper function and not repeating the same logic 3 times?

marusak:
You can maybe try to make it fail? (e.g. break the command on purpose/break networking/break host..., just to make sure it provides a good error messages.
Also make sure it is one commit, these should not be separate :)

dominikvagner (approved):
looking good ✨ hopefully it helps 🙏🏼 
nice job 👏🏼 lgtm! ✅ 

_sidenote: we have a retry helper for general retries of anything, but that doesn't provide the custom output like this and so if we want that for debugging, which could be useful, then this is fine_ 😄

PR #638, opened 2025-09-04
Test: Fix unsafe negative asserts

swadeley:
## Summary

    Negative asserts need additional steps
    before asserting it did not occur.

## Testing steps
tests pass

### DIFF
--- a/_playwright-tests/UI/PopularRepo.spec.ts
+++ b/_playwright-tests/UI/PopularRepo.spec.ts
@@ -133,8 +133,15 @@ test.describe('Popular Repositories', () => {
     await test.step('Shared EPEL repository cannot be edited or deleted', async () => {
       const row = await getRowByNameOrUrl(page, repoName10);
       await row.getByLabel('Kebab toggle').click();
+      await expect(page.getByRole('menu')).toBeVisible();
+      await expect(
+        page
+          .getByRole('menuitem', { name: 'View all snapshots' })
+          .or(page.getByRole('menuitem', { name: 'No snapshots yet' })),
+      ).toBeVisible(); // No snapshot in CI due to time constraints
       await expect(page.getByRole('menuitem', { name: 'Edit' })).not.toBeVisible();
       await expect(page.getByRole('menuitem', { name: 'Delete' })).not.toBeVisible();
+      await expect(page.getByRole('menuitem')).toHaveCount(1);
     });
   });
 });
--- a/_playwright-tests/Integration/AssociatedTemplateCRUD.spec.ts
+++ b/_playwright-tests/Integration/AssociatedTemplateCRUD.spec.ts
@@ -122,14 +122,14 @@ test.describe('Associated Template CRUD', async () => {
         const modal = page.getByRole('dialog');
 
         await expect(
-          modal.getByRole('link', { name: /This template is assigned to \d+ system/i }),
-        ).not.toBeVisible();
-
-        await expect(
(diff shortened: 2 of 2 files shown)


_playwright-tests/UI/SnapshotErrataCountAndFilter.spec.ts
```diff
@@ -115,7 +115,7 @@ test.describe('Snapshot Errata Count and Filter', () => {
       await snapshotListModal.getByRole('button', { name: 'Clear filters' }).click();
       await expect(
         snapshotListModal.getByRole('button', { name: 'Clear filters' }),
-      ).not.toBeVisible();
+      ).not.toBeVisible({ timeout: 1000 });
```
marusak, line 118: I don't think this needs any adjustment. On line 115 you interact with "Clear filters" button, so it is visible and rendered. On line 118 you expect it not to be present. So that is all fine, you are waiting for it disappearing.
(This would be needed to be handled if the whole page would be reloaded, which would be actually a wrong behavior, but I think we are fine)

swadeley: I have reverted that change

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -117,7 +117,7 @@ test.describe('Popular Repositories', () => {
     await test.step('EPEL tab shows only the shared EPEL repository', async () => {
       await page.getByRole('button', { name: 'Custom', exact: true }).click();
       const rows = page.locator('table tbody tr');
-      await expect(rows).toHaveCount(1);
+      await expect(rows).toHaveCount(3);
```
swadeley, line 120: This change was required to get tests to pass locally running against stage and I checked in console (its also has 3 EPEL repos). Waiting to see if CI has same setup.

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -133,6 +133,7 @@ test.describe('Popular Repositories', () => {
     await test.step('Shared EPEL repository cannot be edited or deleted', async () => {
       const row = await getRowByNameOrUrl(page, repoName10);
       await row.getByLabel('Kebab toggle').click();
+      await expect(page.getByRole('menu')).toBeVisible();
```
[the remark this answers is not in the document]
swadeley: That change failed in CI. I have added back the ` await expect(page.getByRole('menu')).toBeVisible();` line.

_playwright-tests/UI/PopularRepo.spec.ts
```diff
@@ -133,8 +133,9 @@ test.describe('Popular Repositories', () => {
     await test.step('Shared EPEL repository cannot be edited or deleted', async () => {
       const row = await getRowByNameOrUrl(page, repoName10);
       await row.getByLabel('Kebab toggle').click();
-      await expect(page.getByRole('menuitem', { name: 'Edit' })).not.toBeVisible();
-      await expect(page.getByRole('menuitem', { name: 'Delete' })).not.toBeVisible();
+      await expect(page.getByRole('menu')).toBeVisible();
+      await expect(page.getByRole('menuitem', { name: 'View all snapshots' })).toBeVisible();
```
xbhouse, line 137: we don't snapshot the epel repo in CI since that would take way too long, so we should only see one item (`No snapshots yet`) in the menu

PR #644, opened 2025-09-09
Build: Fix rename-konflux-pr action

swadeley:
## Summary

Updating title must be done before validating it

## Testing steps
tests pass

### DIFF
--- a/.github/workflows/title.yml
+++ b/.github/workflows/title.yml
@@ -22,6 +22,7 @@ jobs:
 
   title-validate:
     name: Title
+    needs: rename-konflux-pr
     runs-on: ubuntu-latest
     steps:
       - uses: deepakputhraya/action-pr-title@master


.github/workflows/title.yml
```diff
@@ -18,7 +31,17 @@ jobs:
           PR_NUMBER: ${{ github.event.pull_request.number }}
           PR_TITLE: ${{ github.event.pull_request.title }}
         run: |
-          gh pr edit $PR_NUMBER --title "Build: $PR_TITLE"
```
marusak, line 21: https://github.com/content-services/content-sources-frontend/commit/4086622a1abc0d6aae31818e248dcbeb4a47f651

So this patch was not working? Do you have some log where this is failing?

marusak: So it shows green - it failed because it was missing gh command?

marusak:
This seems to make sense, thanks! Do you have link to a successful PR that manage to change the title and then passed also on the title check? Thanks!

dominikvagner (commented):
**question:** did you want to make it so that the `rename-konflux-pr` job runs before the `title-validate` job? 💭 because I think it would make sense for it to run like that, otherwise the title will still fail the first time on konflux PRs 😅

if that's the case, then just putting the rename job before the title validating job in the workflow isn't the solution 🙅🏼  
jobs inside of a workflow are run in parallel, to run them sequentially you need to define dependecies between them, like so:
```yaml
jobs:
  rename-konflux-pr:
    ...
  title-validate:
    needs: rename-konflux-pr
    ...
```

marusak:
> I just trigged this one:
> https://github.com/swadeley/content-sources-frontend/actions/runs/17768149341

But it failed?

I think Dominik might have found the issue :muscle:

.github/workflows/title.yml
```diff
@@ -3,6 +3,11 @@ on:
   pull_request:
     types: [opened, edited, reopened]
 
+permissions:
+  contents: read
```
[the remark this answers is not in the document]
marusak: What about the other line? Have you tried removing that one as well?
I checked logs in https://github.com/content-services/content-sources-frontend/pull/653 and it seems that `pull-request: write` is the default:
https://productionresultssa17.blob.core.windows.net/actions-results/0aea6bcf-4f64-4af8-a7ff-4008627d381a/workflow-job-run-188d2371-6a0d-5bb2-b0b2-4cbaa8bf6329/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-09-17T07%3A05%3A40Z&sig=qj5ytaOrs6bm8TRkBixAfTmxbWs1j%2BeaccB8AquilCw%3D&ske=2025-09-17T17%3A17%3A39Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-09-17T05%3A17%3A39Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-05-05&sp=r&spr=https&sr=b&st=2025-09-17T06%3A55%3A35Z&sv=2025-05-05

dominikvagner (approved):
looks good now! 👍🏼 
ack from me ✅ thanks!

marusak (approved):
If it passed with only the dependency update, than :+1:

PR #649, opened 2025-09-15
Test: Add check URL gets trailing slash

swadeley:
## Summary

    Regression test for HMS-1118
    https://issues.redhat.com/browse/HMS-1118

## Testing steps
tests pass


dominikvagner (approved):
makes sense, thanks! ✅ 🚀

PR #656, opened 2025-09-19
Build: remove patch tests

dominikvagner:
## Summary
This removes Patch related tests as they should live in a different place and are causing test failures because our testing account doesn't have registered systems.


swadeley (approved):
ACK

PR #657, opened 2025-09-19
Build: update fec conf and use new dev cmd

dominikvagner:
## Summary
This updates the "@redhat-cloud-services/frontend-components-config" package to get the new "dev-proxy" command in the FEC binary. The new command should be faster and more reliable for running our frontend in our dev setups (react dev tools will now be more usable when the components aren't minimized).
Also switches our custom "start" commands to use it instead of the older "dev" command.

## Testing steps
Test out our `start` and `local` commands to see if they work as they should. 
_Mind you, if you were running the proxy container standalone, then kill it, it would conflict with this command._


xbhouse:
this just needs a rebase :)

PR #655, opened 2025-09-18
Test: Integration tests after merging the fix for HMS-9325

mayurilahane:
- Add test.use() configuration to automatically authenticate as stable_sam_stage user
- Remove explicit auth headers token in each API request

PR #659, opened 2025-09-22
HMS-9325: update API tests to use test-wide headers

dominikvagner:
## Summary
This updates some API integration tests to use headers set for the entire test/describe, instead of setting them per API call. This was done previously, because of a bug in test-utils due to which the test-wide headers weren't picked up.
Also updates the test-utils submodule.

## Testing steps
Test pass.


mayurilahane:
> i think you and @mayurilahane have both opened PRs fixing the same thing XD (#655)

I see ,
sorry for the misscomunication.

No worries now that this PR already has ACK lets merge this i can close mine 
Thanks !

PR #678, opened 2025-10-13
Test: Disable PW traces for setup

swadeley:
## Summary

    to prevent password logging

## Testing steps
tests pass and PW trace for setup does not show passwords


dominikvagner (approved):
ack ✅ thanks for spotting this! 👏🏼

PR #684, opened 2025-10-14
Test: Fix empty RHEL repo table message

swadeley:
## Summary

The new empty repo table message is context-aware based on the content origin filter. 
When the test filters to show only "Red Hat" repositories, the empty state message would be:
"No Red Hat repositories match your criteria"

## Testing steps
All tests pass, esp. Integration/LayeredRepoAccess.spec.ts


TenSt (approved):
lgtm

PR #689, opened 2025-10-17
Test:  Make the persistent user naming consistent

swadeley:
## Summary

Make the persistent user, stable sam, naming consistent

## Testing steps
API Integration tests pass


dominikvagner (approved):
thanks, this was super confusing, much better now ✨  
ack ✅

PR #700, opened 2025-10-23
Test: Fix template cleanup in Integration tests

swadeley:
## Summary

The template  cleanup helper does not support the use of prefix

## Testing steps
tests pass and templates are not left behind


dominikvagner:
the template cleanup helper [does](https://github.com/content-services/content-sources-backend/blob/42fcb6aa14de3364f4309c7acd3012dc88db0368/_playwright-tests/test-utils/src/helpers/cleanup.ts#L75) support prefixes, it is the way how it determines what templates should be deleted 😅

PR #702, opened 2025-10-24
Test:  Fix persistent repo introspection test

swadeley:
## Summary

   Even unavailable repos must be introspected
   Test must check all repos and then only fail if one or more repos are not being introspected.

## Testing steps
test pass


TenSt (approved):
lgtm

PR #709, opened 2025-10-29
Test: add rhsmClient status helper for associated template tests

swadeley:
## Summary

Add a helper that checks if an rhsmClient is present in inventory and attached to a template
Make use of that helper in Associated Template test

## Testing steps
Integration/AssociatedTemplateCRUD.spec.ts passes

### DIFF
--- a/_playwright-tests/Integration/AssociatedTemplateCRUD.spec.ts
+++ b/_playwright-tests/Integration/AssociatedTemplateCRUD.spec.ts
@@ -133,6 +133,12 @@ test.describe('Associated Template CRUD', () => {
       expect(isAttached, 'system should be removed from inventory').toBe(true);
     });
 
+    // https://issues.redhat.com/browse/HMS-9731: backend cache issue with template deletion
+    await test.step('Reload page to ensure UI reflects system removal', async () => {
+      await page.reload();
+      await expect(page.getByRole('button', { name: 'Create template' })).toBeVisible();
+    });
+
     await test.step('Verify template can now be deleted without warning', async () => {
       const rowTemplate = await getRowByNameOrUrl(page, templateName);
       await rowTemplate.getByLabel('Kebab toggle').click();
--- a/_playwright-tests/Integration/AssociatedTemplateCRUD.spec.ts
+++ b/_playwright-tests/Integration/AssociatedTemplateCRUD.spec.ts
@@ -38,18 +38,23 @@ test.describe('Associated Template CRUD', () => {
 
       await expect(
         page.getByRole('heading', { name: 'Additional Red Hat repositories', exact: true }),
+        'should be on the RHEL repo tab',
       ).toBeVisible();
       await page.getByRole('button', { name: 'Next', exact: true }).click();
 
       await expect(
         page.getByRole('heading', { name: 'Other repositories', exact: true }),
+        'should be on the Other repositories tab',
       ).toBeVisible();
       await page.getByRole('button', { name: 'Next', exact: true }).click();
 
       await page.getByText('Use the latest content', { exact: true }).click();
       await page.getByRole('button', { name: 'Next', exact: true }).click();
 
-      await expect(page.getByText('Enter template details')).toBeVisible();
+      await expect(
+        page.getByText('Enter template details'),
+        'should be on the Enter template details tab',
+      ).toBeVisible();
       await page.getByPlaceholder('Enter name').fill(`${templateName}`);
       await page.getByPlaceholder('Description').fill('Template test for associated system CRUD');
       await page.getByRole('button', { name: 'Next', exact: true }).click();
@@ -57,7 +62,9 @@ test.describe('Associated Template CRUD', () => {
       await page.getByRole('button', { name: 'Create other options' }).click();
       await page.getByText('Create template only', { exact: true }).click();
       const rowTemplate = await getRowByNameOrUrl(page, `${templateName}`);
-      await expect(rowTemplate.getByText('Valid')).toBeVisible({ timeout: 660000 });
+      await expect(rowTemplate.getByText('Valid'), 'repo should show Valid status').toBeVisible({
+        timeout: 660000,
+      });
     });
 
     await test.step('Register system with template using RHSM client', async () => {
@@ -72,7 +79,7 @@ test.describe('Associated Template CRUD', () => {
         console.log('Registration stdout:', reg?.stdout);
         console.log('Registration stderr:', reg?.stderr);
       }
-      expect(reg?.exitCode).toBe(0);
+      expect(reg?.exitCode, 'registration should be successful').toBe(0);
 
       await refreshSubscriptionManager(regClient);
     });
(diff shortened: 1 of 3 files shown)


_playwright-tests/Integration/AssociatedTemplateCRUD.spec.ts
```diff
@@ -74,9 +77,15 @@ test.describe('Associated Template CRUD', async () => {
       await refreshSubscriptionManager(regClient);
     });
 
+    await test.step('Verify system is attached to template', async () => {
+      hostname = await regClient.GetHostname();
+      console.log('System hostname:', hostname);
```
marusak: Is this debugging or we want to keep it in? :thinking:

_playwright-tests/Integration/AssociatedTemplateCRUD.spec.ts
```diff
@@ -74,9 +77,15 @@ test.describe('Associated Template CRUD', async () => {
...
+      const isAttached = await pollForSystemTemplateAttachment(page, hostname, 10_000, 6);
+      expect(isAttached).toBe(true);
```
marusak: I just learned the other day that "expect" can have second argument. See https://playwright.dev/docs/test-assertions#custom-expect-message

That is cool, because having these bare "toBe(true)" show up in logs just as "toBe(True)" which does not say a lot. This step is very simple, it has one assert so it would be straight forward what failed, but may be a good practice to start adding messages to these asserts

_playwright-tests/Integration/helpers/systemHelpers.ts
```diff
@@ -0,0 +1,92 @@
...
+
+      // Find the system with matching host name
+      const system = body.data.find(
+        (sys: { attributes: { display_name: string } }) => sys.attributes.display_name === hostname,
+      );
+
+      if (!system) {
+        // System not found in inventory; it should not be attached to a template
+        console.log(`System '${hostname}' not found in inventory (not attached)`);
+        return false;
```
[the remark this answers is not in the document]
swadeley: Ok, good point, will change to poll and log attempts same as the other steps.

_playwright-tests/Integration/AssociatedTemplateCRUD.spec.ts
swadeley: Ok, sounds good, I will do that when other changes have been made.

_playwright-tests/Integration/AssociatedTemplateCRUD.spec.ts
```diff
@@ -74,9 +77,15 @@ test.describe('Associated Template CRUD', async () => {
...
+    });
+
     await test.step('Attempt to delete template and verify warning appears', async () => {
       await navigateToTemplates(page);
-
```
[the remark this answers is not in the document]
swadeley: Hi, there is a commit to fix that in older PR: https://github.com/content-services/content-sources-frontend/pull/696/commits/2cf8fe62852b990ffe3b449a21978705f7aff5e3

_playwright-tests/Integration/AssociatedTemplateCRUD.spec.ts
```diff
@@ -107,6 +116,13 @@ test.describe('Associated Template CRUD', async () => {
       expect(unreg?.exitCode).toBe(0);
     });
 
+    await test.step('Wait for system to be removed from inventory', async () => {
+      // Poll until system is not found or not attached to template
+      // This should ensure the backend has processed the unregistration before checking UI
+      const isAttached = await pollForSystemTemplateAttachment(page, hostname, 10_000, 6);
```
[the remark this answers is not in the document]
swadeley: I will wait for you further review.

_playwright-tests/Integration/AssociatedTemplateCRUD.spec.ts
```diff
@@ -104,38 +122,56 @@ test.describe('Associated Template CRUD', async () => {
...
           modal.getByText(
             `Template ${templateName} and all its data will be deleted. This action cannot be undone.`,
           ),
+          'delete template modal body should be visible',
         ).toBeVisible();
 
         await expect(
           modal.getByRole('link', { name: /This template is assigned to \d+ system/i }),
+          'system assignment link should not be visible',
         ).not.toBeVisible();
```
katarinazaprazna: I've recently learned that the direct opposite of `.toBeVisible()` is `.toBeHidden()`. We could try using `.toBeHidden()` as it is more descriptive of the intent in the code

_playwright-tests/Integration/AssociatedTemplateCRUD.spec.ts
```diff
@@ -104,38 +122,56 @@ test.describe('Associated Template CRUD', async () => {
...
+        await expect(removeButton, 'delete button should be enabled').toBeEnabled();
         await removeButton.click();
       });
 
       await test.step('Verify template is removed from the list', async () => {
-        await expect(rowTemplate.getByText('Valid')).not.toBeVisible({ timeout: 30000 });
+        await expect(
+          rowTemplate.getByText('Valid'),
+          'template should not be visible in the list',
+        ).not.toBeVisible({ timeout: 30000 });
```
katarinazaprazna: Could we apply the same suggestion as mentioned in my comment above regarding the `.toBeHidden()` matcher?

katarinazaprazna (changes requested):
I've completed my review and left a few comments outlining some minor suggestions. Overall, the changes look great! Thanks for putting this together :)

@marusak, would you be able to provide a quick re-review? Thanks so much!

_playwright-tests/Integration/helpers/rhsmClient.ts
```diff
@@ -193,16 +205,37 @@ export class RHSMClient {
   async Destroy(unregisterMethod: 'rhc' | 'sm' | 'none' = 'none') {
```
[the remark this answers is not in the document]
swadeley: Ok, changes made. Thank you.

_playwright-tests/Integration/AssociatedTemplateCRUD.spec.ts
swadeley: Use negative matcher where possible #733

_playwright-tests/Integration/AssociatedTemplateCRUD.spec.ts
```diff
@@ -107,6 +116,18 @@ test.describe('Associated Template CRUD', () => {
...
     });
 
+    await test.step('Wait for system to be removed from inventory', async () => {
+      // Poll until system is not found or not attached to template
+      // This should ensure the backend has processed the unregistration before checking UI
+      const isAttached = await pollForSystemTemplateAttachment(page, hostname, false, 10_000, 6);
+      expect(isAttached).toBe(true);
+    });
+
+    await test.step('Reload page to ensure UI reflects system removal', async () => {
```
marusak: Ugh, do we really need to reload the page?

marusak:
Looking good. One note, one question (or maybe I am just confused about usage of .not.toBeEnabled()). Mainly though you added page reloading. That is a big anti-pattern. The whole point about not using `NavigateTo...()` multiple times was to avoid reloading pages. Do we have to do it as the page never updates? If so, then that is a bug. (We can report it and add note why we are doing this workaround)

_playwright-tests/Integration/AssociatedTemplateCRUD.spec.ts
swadeley: Fixed that, thank you

swadeley: I decided to use `toHaveCount(0)`

marusak (approved):
Thank you!

katarinazaprazna (approved):
Thank you so much for taking the time to make these thoughtful changes! ✅

PR #761, opened 2025-11-24
Test: Add token refresh helper

swadeley:
## Summary

  To enable tokens to be refreshed at test start.

## Testing steps

#testwith https://github.com/content-services/content-sources-backend/pull/1310

### DIFF
--- a/_playwright-tests/helpers/tokenHelpers.ts
+++ b/_playwright-tests/helpers/tokenHelpers.ts
@@ -1,0 +1,210 @@
+import { type Page } from '@playwright/test';
+import path from 'path';
+import { readFileSync, existsSync } from 'fs';
+
+// Type definitions
+interface JWTPayload {
+  exp: number;
+  iat: number;
+  [key: string]: unknown;
+}
+
+interface AuthCookie {
+  name: string;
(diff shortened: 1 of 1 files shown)


dominikvagner (changes requested):
this really shouldn't split between repos like this :/
it's very unusual, not a good practice and bound to cause issues 🙅🏼 [[this](https://github.com/content-services/content-sources-backend/pull/1310/files#diff-4cbb11af18195a540a2aebd0fd2c57adeb05bd5211c5d3c802d16efaec872cb6R43-R57)]
this could very well just live in the `test-utils` in our backend repo and there shouldn't be an issue with it, if it is made to just ignore the refreshing when there isn't a browser state saved ✍🏼 
please close this PR and move this to the utils so it lives alongside the fixture 🔧🚚  

_more of a though then a strict change request 💭: it would be nice if it could be simplified a bit and less AI generated (the console logging could be cut down significantly, it could be less defensive and those unnecessary comments in code could be reduced, function descriptions are perfectly acceptable) 😅_

swadeley:
Moving to test-utils submodule in backend repo.

PR #763, opened 2025-11-24
HMS-9590: refactor create and edit template worflow

Starle21:
## Summary - CREATE AND EDIT TEMPLATE WORKFLOW REFACTOR

The goal is to improve readability, modularity and changeability of the create and edit template workflow. The overall aim is to make functionality explicit and easy to change. This can be achieved by making each step independent on the others. Each step then encapsulates domain logic (core, use-cases), api requests and UI.

### Problems

Files in the Pages folder are a ball of mud. They are an incoherent mesh of code. They try to combine multiple use-cases, their data transformations and derived data for a pile of UI components each of them gathers. They are written as one giant component, easily reaching 500 lines of code. There is a page (ContentListTable), which reaches 930 lines. AddOrEditTemplate folder, where the create and edit template workflow resides, is no exception.

Not only it is bad from a performance perspective - when there is any state change, a whole giant component has to rerender, it is hard to understand where one functionality ends and another starts. It is hard to understand what a particular piece of code does. Which use-case does it belong to? There is no `separation of concerns`. Each time we want to change a detail on a page of any step, we have to understand and go through all of it. We have to mentaly group together the parts that go together. It would be better to make this explicit, to define separate functions and components for these groups once so we don't have to keep doing this again and again every time we work with a particular Page.

### Remedies

In general, there are 2 concepts used to fix the ball of mud - `separation of concerns` and `put together what changes together (coherence)`.

1. In UI, big components are split into constituent subcomponents. These subcomponents read the logic that concerns only them from a context store that each step has. Early returns help one see explicitly what versions a component can be in, depending on a state.

2. Logic is taken out of UI and placed into a particular context store of each step. Components only read and update data through their store api. If there was a change in logic in future, it would only have to change in a store, the api would stay the same and no change would have to be done in consuming components.

3. Business processes are translated in software as use-cases. They are the main reason any business application exists. Multiple functions, spread over multiple useEffects and setStates (as in DefineContentStep or SetUpDateStep) has been used up until now. Each use-case is now encapsulated in one function (chooseHardcodedUUIDs, validateDependencies). It groups multiple steps in a data transfomation pipeline, where only the ending state is the one that is saved.

4. Hooks are used as a composition tool, so use-case functions can stand alone, but at the same time communicate with the rest of the workflow, read and set state in the top level TemplateStore context.

5. To make the use-cases and domain entities explicit, extensive typing is employed. Types are defined using type aliases. Types define domain, both data in its various states (TemplateRequestStartStep, TemplateRequestFurtherSteps, TemplateRequestFinalized) and data transformations (use-cases: CreateNewTemplate, ToggleSelectedOtherRepository). Compound types are defined by ANDing or ORing simple types (so called algebraic types).

6. Ports files make explicit the most important functions on the "edges" of a domain core. Input ports define the shape of data that a domain core accepts and which functionalities it provides to UI. Output ports declare what other services the core relies on, without which it cannot function.

7. Related code is grouped into one folder (feature) - application core with domain types, domain functions and use-cases; client store; api and UI. It is a so-called vertical feature slice, that cuts through multiple technical layers (api, runtime store, domain logic), that are usually always touched when we need to make a change. It is collocated together, to improve coherence of the workflow.

8. When multiple steps use the same function, it is extracted into the shared folder. It is better to make sure components from multiple steps don't import functions directly from one another, which would start to create a ball of mud again. Workflow folder with TemplateStore can be considered as being "on top" of other steps that communicate with it through context stores or in use-case composition hooks. Otherwise, workflow only composes the overall create or edit workflow steps. EditTemplate feature is a similar to workflow, that it is on top of other steps and they read directly from it.

9. If a whole application is structured with vertical feature slices, outside of them there is only a generic and reusable code left. We can define reusable hooks, UI components without any business logic (presentational components), api service, generic react context runtime store or persistence service. If an api or UI code is particularly coupled with a feature logic, it is put under the feature slice and not in generic services.

### Commits and phases in the PR

To help with reviewing the PR, there can be 4 phases distinguished in the commits:

0. the first 5 commits are preparation steps.
1. start of the refactor - `move: AddOrEditTemplate under features`
2. extracting logic out of UI into stores or into standalone functions for each step. UI components are split into subcomponents.
3. end of the 1. phase - `rename: AddOrEditTemplate -> TemplateModalBase`
4. extracting logic that does not fit in the top level context
5. end of the 2. phase - `extract: queryClient from AddTemplateContext`
6. changing storage in the top level context and the related apis. only each step stores communicate with the top level context, components communicate with their respective store.
7. end of the 3. phase - `rename: AddTemplateContextProvider -> TemplateStore`
8. extracting use-cases out of stores into standalone functions. Stores compose these functions and provide them to components in their step.
9. end of the 4. phase and of the refactor - `extract, change: structure of CustomRepositoriesStore`
Please open and see detailed notes for the refactor in `src/features/notes-refactor`.

### 
If you feel like you want to learn more about the concepts, you can check out documentation in the `src/features/notes-refactor` folder. However, it is not at all necessary and I will remove it before merging.

### Manual testing
Run either against stage or against locally running backend. Try out both create and edit workflows from start to end to see that either a new template is created or a change to an already existing template is made.

### Playwright tests

For testing with playwright locally, you need to run the backend PR below. `make repos-minimal` command imports extra repositories that the template workflow need to run correctly:

#testwith https://github.com/content-services/content-sources-backend/pull/1358


dominikvagner (commented):
first off: wow! 😮 that's a lot of work! thank you!! 👏🏼😌 
_(the commit splitting, makes this sort of reviewable and shows of the steps, thx! although still not line by line reviewable so we will have to trust our tests and manual testing 😄)_

never seen or worked with a frontend code written like this, so there is a lot to learn and get used to 🤓 
the vertical/feature slice approach seems quite cool! time will tell how nice it is to work in, but in general it looks good, and I have high hopes! **let's try it out!** ✨ 

---
I am not the first to review this and have to agree with Kate and Bryttanie on most of the stuff they pointed out 💯 
 - _all the amazing stuff_ 🤩: **decoupling/grouping**, UX fixes like the warning, extracting logic out of components, **early returns**, possibly better maintainability in the future  
 - _and the worrying_ 😨: complexity (context hell, amount of abstractions, learning curve), maintenance of both structures (<- this might be my biggest concern), too many folders (and new names 😆)
 
you know all the pros and why this is nice, that's why you did this 😄 
so here are a few comments on the cons from someone unfamiliar _(and maybe some suggestions how to make this more digestible and easier to slot into the project as a whole?)
(not all of these are necessary change requests, just my opinions, some of which might be caused by my inexperience/unfamiliarity)_ 💭:
- contexts being called `SomethingApi` or `Something[Derived]State` can be a bit confusing, I still don't really get the difference
- too many folders and no "entry point"
   - this made me a bit lost initially, when you open the feature, workflow or step directories you only get folders, and the steps have `core` and `ui`, this prompts a question where should I start looking first? 😵‍💫 
   - I think "exploding" the UI folder would help, i.e. the `SomethingStep[.test].tsx` files would be at the root of the step folder and there wouldn't be an `ui` folder, just a `components` folder 💭 this would help with navigating the code imho. \
    _(and also maybe pulling the insides of `ui` from `workflow` into the root of `createAndEditTemplate`?)_
   - and/or "exploding" the core folders (so that ports and types are at the root) and renaming domain to helpers _(I think you even mentioned doing that rename)_
- rename `FirstEmpty` --> `Option|Optional`, that might be clearer/more explicit
- really a nitpick: rename `DependencyNotfication` and subsequent hooks, etc. to something like `Dependency[Status|Warning|Banner]`? we mostly use the name notification for toast notifications, so this could be clearer what's what 😅 
- I would be cautious with the amount of stores, there is quite a lot of them (ex.: repo steps), which creates some complexity imo 😵‍💫
   - as Kate mentioned: _"bit of light coupling can be helpful for readability and DX ... just so we don't inadvertently increase cognitive load"_
   which I think might be the case here (or getting really close to) so that we don't have a context hell (and possibly in other places)

---
some more on-the-go thoughts, while reviewing commit by commit, in other comments 💭 

great work! 💪🏼👏🏼

PR #777, opened 2025-12-04
HMS-9584: Add test to verify cleanup function of expired snapshots

mayurilahane:
## Summary

- Check for expired snapshots in Red Hat and EPEL repositories
- Verify stage(91 days) and prod(366 days) environments
- Make sure to have at least 1 snapshot in the repo

### DIFF
--- a/_playwright-tests/Integration/CheckSnapshotExpiry.spec.ts
+++ b/_playwright-tests/Integration/CheckSnapshotExpiry.spec.ts
@@ -1,0 +1,79 @@
+import { test, expect } from 'test-utils';
+import {
+  RepositoriesApi,
+  SnapshotsApi,
+  ListRepositoriesRequest,
+  ListSnapshotsForRepoRequest,
+  ApiSnapshotResponse,
+} from 'test-utils/client';
+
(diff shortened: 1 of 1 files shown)


TenSt (dismissed):
lgtm

_playwright-tests/Integration/CheckSnapshotExpiry.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+
+        if (snaps.data!.length === 1 && expired.length === 1) {
+          console.log(
+            `Repo ${repo.name} has only 1 snapshot (expired). Skipping - repo must keep at least 1 snapshot.`,
+          );
+          continue;
+        }
+
+        console.log(
+          `Expired snapshots in repo: ${repo.name} (UUID: ${repo.uuid}). Created at: ${expiredDates}`,
```
rverdile: We'll see what repos have expire snapshots when the test fails, so I don't think this log is needed

rverdile (commented):
just some comments about logging. otherwise this should be good once the tests can pass

PR #780, opened 2025-12-05
Test: Wait for valid repo before uploading

swadeley:
## Summary
  To aid test stability.
  https://issues.redhat.com/browse/HMS-9856
## Testing steps
Integration/InstallUploadRepoContent  passes

### DIFF
--- a/_playwright-tests/UI/UploadRepo.spec.ts
+++ b/_playwright-tests/UI/UploadRepo.spec.ts
@@ -1,5 +1,5 @@
 import path from 'path';
-import { test, expect, cleanupRepositories } from 'test-utils';
+import { test, expect, cleanupRepositories, waitWhileRepositoryIsPending } from 'test-utils';
 import { navigateToRepositories } from './helpers/navHelpers';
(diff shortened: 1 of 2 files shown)


TenSt (approved):

xbhouse:
i think we'd want to wait for the repo to be valid in the integration test (not the UI test) and wait for the upload repo to be created first

_playwright-tests/UI/UploadRepo.spec.ts
```diff
@@ -22,6 +22,9 @@ test.describe('Upload Repositories', () => {
       const nameInput = page.getByPlaceholder('Enter name');
       await nameInput.click();
       await nameInput.fill(uploadRepoName);
+      // Upload can fail if repository is not valid HMS-9856
+      const newRepoRow = await getRowByNameOrUrl(page, uploadRepoName);
+      await expect(newRepoRow.getByText('Valid')).toBeVisible();
```
xbhouse: this is before the repo is created so this will fail, we'd need to wait for the status to be valid after bulk_create is called. similar to where you added the check in the integration test

xbhouse (approved):
with the polling it passes consistently for me locally :) lgtm! thank you!

PR #825, opened 2026-01-12
Test: Make token and auth file names match

swadeley:
## Summary

To avoid using maps to translate between them.

## Testing steps
Tests pass


dominikvagner (commented):
this makes sense: `TOKEN` to `ADMIN_TOKEN` 👍🏼😄
for the filenames: I think we could just name them exactly like the env vars, i.e.:  `readOnlyToken.json` -> `READ_ONLY_TOKEN.json`, that way we wouldn't have to do any parsing on the backend (except trimming the file extension), what do you think?

dominikvagner (approved):
ack! ✅ thanks! 😇

PR #871, opened 2026-02-12
Test: Set dependabot ignore for major updates to eslint

swadeley:
## Summary

 Set dependabot ignore for major updates to eslint

Hold Eslint version to 9.x because with ESLint 10, eslint-plugin-react breaks for every file
(e.g. src/App.tsx) because of the context.getFilename API change, not only for Playwright tests.

## Testing steps
tests pass


dominikvagner (approved):
ack! ty! ✅

PR #899, opened 2026-03-10
Test: fix errata check to poll for updates

swadeley:
## Summary

As we expect the errata count to change we should poll for that change.

## Testing steps

Integration tests pass


katarinazaprazna (approved):
Thank you!

PR #941, opened 2026-03-30
Test: Use currents key from codebuild

swadeley:
## Summary
Use currents key from codebuild
Use [currents.dev](https://app.currents.dev/) project for "Repos - Frontend"

## Testing steps
UI project tests get upload to currents.dev


xbhouse (approved):
thank you! UI test results are back in currents 🥳 

i think we **could** add the `CURRENTS_PROJECT_ID` to codebuild and pull it from there, but it looks like we haven't done that previously and don't do this for integration tests, so i won't block on that

PR #948, opened 2026-03-31
HMS-10325: add minor version and release stream to templates table

xbhouse:
## Summary

- Adds minor OS version data to the templates table
- Release stream is added as a label next to the template name
- Adds support for filtering templates by multiple major and minor versions
- Adds support for filtering templates by release streams
  
Screenshot From 2026-04-21 11-46-03
<br>
Screenshot From 2026-04-21 11-46-41

## Testing steps

1. Create templates of various OS versions and release streams. These values should be shown correctly in the OS version column and in a label next to the template name
2. Filtering by minor and major OS versions should work as expected
3. Filtering by release stream should work as expected
4. Tests pass

### DIFF
--- a/src/Pages/Templates/TemplatesTable/TemplatesTable.tsx
+++ b/src/Pages/Templates/TemplatesTable/TemplatesTable.tsx
@@ -4,6 +4,7 @@ import {
   FlexItem,
   Grid,
   Icon,
+  Label,
   Pagination,
   PaginationVariant,
   Spinner,
@@ -45,6 +46,7 @@ import { ExclamationTriangleIcon, ExternalLinkSquareAltIcon } from '@patternfly/
 import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 import { useFlag } from '@unleash/proxy-client-react';
 import { TEMPLATES_DOCS_URL } from './constants';
+import { abbreviateStreamName } from './helpers';
 
 const useStyles = createUseStyles({
   topContainer: {
@@ -91,10 +93,12 @@ const TemplatesTable = () => {
 
   const defaultValues: TemplateFilterData = {
     arch: '',
-    version: '',
+    version: [],
+    extended_release_version: [],
     search: '',
     repository_uuids: '',
     snapshot_uuids: '',
+    extended_release: [],
   };
 
   const [filterData, setFilterData] = useState<TemplateFilterData>(defaultValues);
@@ -102,9 +106,13 @@ const TemplatesTable = () => {
   const clearFilters = () => setFilterData(defaultValues);
 
   const notFiltered =
-    filterData.arch === '' && filterData.version === '' && filterData.search === '';
+    filterData.arch === '' &&
+    filterData.version.length === 0 &&
+    filterData.extended_release_version.length === 0 &&
+    filterData.search === '' &&
+    filterData.extended_release.length === 0;
 
-  const columnSortAttributes = ['name', '', 'arch', 'version', '', ''];
+  const columnSortAttributes = ['name', '', 'arch', 'version', '', '', ''];
 
   const sortString = useMemo(
     () =>
@@ -189,6 +197,8 @@ const TemplatesTable = () => {
     isError: repositoryParamsIsError,
     getArchName,
     getVersionName,
+    getStreamName,
+    getMinorVersionName,
   } = useDistributionDetails();
 
   const actionTakingPlace = isLoading || isFetching || repositoryParamsLoading;
@@ -313,6 +323,8 @@ const TemplatesTable = () => {
                       last_update_snapshot_error,
                       last_update_task,
                       to_be_deleted_snapshots,
+                      extended_release,
+                      extended_release_version,
                     }: TemplateItem,
                     index,
                   ) => (
@@ -327,6 +339,11 @@ const TemplatesTable = () => {
                               {name}
                             </Button>
                           </FlexItem>
+                          {extended_release ? (
+                            <Label isCompact color='blue'>
+                              {abbreviateStreamName(getStreamName(extended_release))}
+                            </Label>
+                          ) : null}
                           <Hide
                             hide={!(to_be_deleted_snapshots && to_be_deleted_snapshots.length > 0)}
                           >
@@ -348,7 +365,11 @@ const TemplatesTable = () => {
                       </Td>
                       <Td>{description}</Td>
                       <Td>{getArchName(arch)}</Td>
-                      <Td>{getVersionName(version)}</Td>
+                      <Td>
+                        {extended_release && extended_release_version
+                          ? getMinorVersionName(extended_release_version)
+                          : getVersionName(version)}
+                      </Td>
                       <Td>
                         <ConditionalTooltip
                           show={!use_latest}
--- a/src/Pages/Templates/TemplatesTable/components/TemplateFilters.tsx
+++ b/src/Pages/Templates/TemplatesTable/components/TemplateFilters.tsx
@@ -12,11 +12,13 @@ import {
   MenuToggle,
   DropdownList,
   DropdownItem,
+  TreeView,
+  TreeViewDataItem,
 } from '@patternfly/react-core';
 
 import { FilterIcon, SearchIcon } from '@patternfly/react-icons';
 import Hide from 'components/Hide/Hide';
-import { RepositoryParamsResponse } from 'services/Content/ContentApi';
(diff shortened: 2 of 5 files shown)


src/Pages/Templates/TemplatesTable/components/TemplateFilters.tsx
```diff
@@ -223,18 +401,62 @@ const Filters = ({ isLoading, setFilterData, filterData }: Props) => {
...
+      case 'Release stream':
+        return (
+          <Dropdown
+            onSelect={(_, val) => {
+              setSelectedStreams((prev) =>
+                prev.includes(val as string)
+                  ? prev.filter((item) => item !== (val as string))
+                  : [...prev, val as string],
+              );
+            }}
```
katarinazaprazna, line 438: One small suggestion. Since this is a multi-select where we want to ensure uniqueness, what do you think about using a Set? It would help future-proof the code against accidental duplicates

katarinazaprazna (dismissed):
Awesome work! ✨ I added a minor non-blocking suggestion, but it's purely a nice-to-have

src/Pages/Templates/TemplatesTable/components/TemplateFilters.tsx
```diff
@@ -107,28 +157,172 @@ const Filters = ({ isLoading, setFilterData, filterData }: Props) => {
   useEffect(() => {
     setFilterData({
       search: debouncedSearchQuery,
-      version: getLabels('version', debouncedSelectedVersion),
+      version: getLabels('version', debouncedSelectedMajorVersion),
```
katarinazaprazna: Hey @xbhouse, I noticed that with the addition of minor versions, the column sorting logic seems to have broken. I’m running into the same issue while trying to sort the repositories table by OS version

I'll go ahead and create a BE ticket for this

Image

CC: @rverdile

src/Pages/Templates/TemplatesTable/components/TemplateFilters.tsx
xbhouse: great point! using a set now for streams and minor versions

src/Pages/Templates/TemplatesTable/components/TemplateFilters.tsx
rverdile, line 1: I'm wondering if it makes sense to have a "release_stream" column. A user with no extended support subscription never actually sees a "release stream" during creation, but would still see this column. And that column would then have no real value. Maybe a blue chip next to the name would be a better way to show the release stream. What do you think?

src/Pages/Templates/TemplatesTable/TemplatesTable.tsx
```diff
@@ -92,19 +93,25 @@ const TemplatesTable = () => {
...
+    filterData.arch === '' &&
+    filterData.version === '' &&
+    filterData.extended_release_version.length === 0 &&
+    filterData.search === '' &&
+    filterData.extended_release.length === 0;
 
-  const columnSortAttributes = ['name', '', 'arch', 'version', '', ''];
+  const columnSortAttributes = ['name', '', 'arch', 'version', '', '', '', ''];
 
   const sortString = useMemo(
```
rverdile, line 117: It looks like sorting by version may not work correctly with the minor versions since it'd only be sorting on the major version column (make sure names are not coincidentally alphabetical). I think it would be enough to just sort by both columns, but that would have to be added to the backend. You could send the correct sort parameters here though and we can fix it after.

rverdile (commented):
I was playing around with the PR a little, two comments. Open to everyone's thoughts

src/Pages/Templates/TemplatesTable/TemplatesTable.tsx
rverdile: sort by both columns meaning the the API request would look like `sort_by=version:asc,extended_release_version:asc`

src/Pages/Templates/TemplatesTable/components/TemplateFilters.tsx
katarinazaprazna: Good eye on the subscription guard! If we stick with the column, hiding it for non-subscribers makes total sense, though I'm also liking the chip idea

I know Maria has been in discussions with the design team about EUS and our table layouts to keep everything consistent. It might be worth looping her in for a quick opinion. What do you all think?

src/Pages/Templates/TemplatesTable/TemplatesTable.tsx
katarinazaprazna: Here's a ticket to track this: https://redhat.atlassian.net/browse/HMS-10532. I’ve got it assigned to myself for the time being

xbhouse: good catch! i think we'd still need to add `extended_release_version` [here](https://github.com/content-services/content-sources-backend/blob/b94464d768783429da2f3010073af83d4c224aa1/pkg/dao/templates.go#L335) before sending it through from the frontend. without that it'll throw a 500. 

i can open a ticket to support sorting by both versions and pick that up. until that's implemented though, we'd either have broken sorting or need to make this column not sortable for a little bit. what do you think?

src/Pages/Templates/TemplatesTable/components/TemplateFilters.tsx
xbhouse: that's a good point. i think a chip or label could work well. do you mean similar to what we have for upload and community repos?

rverdile: yeah exactly, something like that

src/Pages/Templates/TemplatesTable/TemplatesTable.tsx
rverdile: oh okay, I thought it would ignore an unused parameter. I'd say let's leave it broken under the assumption that we're about to fix it, or if you prefer just go fix it first

rverdile: I missed this but it looks like @katarinazaprazna opened a ticket: https://redhat.atlassian.net/browse/HMS-10532

xbhouse: oh awesome, thanks @katarinazaprazna!  :rocket:

katarinazaprazna:
Hi @xbhouse and @rverdile. It looks like my previous comments might not have posted correctly

I wanted to let you know that I’ve opened [HMS-10532](https://redhat.atlassian.net/browse/HMS-10532) to make sure that the sorting issue is tracked. I've assigned it to myself for now, but if anyone else has a preference for taking this one on, feel free to grab it

Screenshot From 2026-04-15 21-29-01

[HMS-10532]: https://redhat.atlassian.net/browse/HMS-10532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

src/Pages/Templates/TemplatesTable/components/TemplateFilters.tsx
xbhouse: just pushed that up!

src/Pages/Templates/TemplatesTable/components/TemplateFilters.tsx
```diff
@@ -107,28 +158,196 @@ const Filters = ({ isLoading, setFilterData, filterData }: Props) => {
...
+                })),
+                defaultExpanded: true,
+              }
+            : {}),
+        };
+      }),
+    [distribution_versions, minorsByMajor, selectedMajorVersions, selectedMinorVersions],
+  );
+
+  // version filter helpers
```
xbhouse, line 276: @katarinazaprazna @rverdile just pushed up the revised version filtering. it's more complex but i tried to cover all the edge cases, let me know if you see anything weird 👀 

also @katarinazaprazna i just saw we have a finalized design so i'll work on updating this to match that

xbhouse: i opened this [backend PR](https://github.com/content-services/content-sources-backend/pull/1469) to try to simplify the filtering for templates and repositories, so this is dependent on that. i'll put this back in draft for now

katarinazaprazna (dismissed):
Many thanks for tackling this one! It looks great 🧡

katarinazaprazna (approved):

PR #985, opened 2026-04-28
HMS-10505: filter repo table by minor version

katarinazaprazna:
## Summary

- Replace the OS version checkbox filter with `DataViewTreeFilter` that supports both major versions (e.g., RHEL 9) and minor/EUS versions (e.g., RHEL 9.6)
- Bump DataView to 6.4.0-prerelease.12 (required for `DataViewTreeFilter` support)
- Migrate all DataView imports to `/dist/dynamic/` paths for proper tree-shaking
- Rename the operating system column header to "OS" and update the filter label to "Operating system" to stay consistent with other Insights apps like Inventory and Patch

Screenshot From 2026-04-28 21-21-16

## Testing Steps
I recommend testing the frontend against stage, as it already contains EUS repositories. This saves you the trouble of manually importing EUS repos, which would be necessary if running against a local backend

* Login: Use the EUS test account credentials
* Navigate to the Repositories table
* Test various combinations of OS version filtering (both major and minor versions)
* Verify that sorting the OS column works correctly in both ascending and descending order

### DIFF
--- a/src/Pages/Repositories/ContentListTable/ContentListTable.tsx
+++ b/src/Pages/Repositories/ContentListTable/ContentListTable.tsx
@@ -1,7 +1,6 @@
 import {
   DataViewToolbar,
   DataViewTable,
-  DataViewTh,
   DataView,
   DataViewState,
   useDataViewSort,
@@ -9,6 +8,7 @@ import {
   DataViewTextFilter,
   DataViewCheckboxFilter,
   DataViewTrObject,
+  DataViewTreeFilter,
 } from '@patternfly/react-data-view';
 import { ThProps, ActionsColumn, IAction } from '@patternfly/react-table';
 import {
@@ -59,10 +59,11 @@ import { DataViewFilters } from '@patternfly/react-data-view/dist/dynamic/DataVi
 import { useContentListFilters, FilterLabelsMap } from './hooks/useContentListFilters';
 import ContentOriginFilter from './components/ContentOriginFilter';
 import CommunityRepositoryLabel from '../../../components/RepositoryLabels/CommunityRepositoryLabel';
-import { DataViewTr } from '@patternfly/react-data-view/src/DataViewTable';
+import { DataViewTr, DataViewTh } from '@patternfly/react-data-view/src/DataViewTable';
 import EmptyTableDataView from 'components/EmptyTableDataView/EmptyTableDataView';
 import CustomEpelWarning from 'components/RepositoryLabels/CustomEpelWarning';
 import { isEPELUrl } from 'helpers';
+import { hasOrigin, versionNameToApiValue } from '../helpers';
 
 type ActionRowData = Pick<
   ContentItem,
@@ -83,9 +84,6 @@ const useStyles = createUseStyles({
 });
 
 export const perPageKey = 'contentListPerPage';
-
-const hasOrigin = (value: unknown): value is { origin?: ContentOrigin } =>
-  typeof value === 'object' && value !== null && 'origin' in value;
 
 interface ContentInformationCellProps {
   rowData: Pick<ContentItem, 'name' | 'url' | 'last_snapshot' | 'origin'>;
@@ -188,6 +186,7 @@ const ContentListTable = () => {
   const {
     getArchName,
     getVersionName,
+    getMinorVersionName,
     isError: repositoryParamsIsError,
     error: repositoryParamsError,
   } = useDistributionDetails();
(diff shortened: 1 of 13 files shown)


xbhouse:
https://issues.redhat.com/browse/HMS-10505

src/Pages/Repositories/ContentListTable/ContentListTable.tsx
```diff
@@ -781,13 +787,24 @@ const ContentListTable = () => {
...
-                  options={osFilterOptions}
+                  // TODO: Create an issue in the data-view repository to support a custom aria-label and a placeholder
+                  // aria-label='filter by operating system'
+                  // placeholder='Filter by operating system'
+                  defaultExpanded={true}
+                  items={osFilterOptions}
+                  value={filters.versions.map((version) =>
+                    version.includes('.') ? getMinorVersionName(version) : getVersionName(version),
+                  )}
+                  onChange={(_event, selectedNames) => {
```
xbhouse: small suggestion, maybe `selectedNames` should be renamed to `selectedVersions` or `selectedVersionNames` just for clarity? 

also to avoid an `any` type for this, i think typing explicitly could work here

src/Pages/Repositories/ContentListTable/ContentListTable.tsx
```diff
@@ -781,13 +787,24 @@ const ContentListTable = () => {
                   placeholder='Filter by name/url'
                   isDisabled={isLoading}
                 />
-                <DataViewCheckboxFilter
+                <DataViewTreeFilter
```
xbhouse, line 790: i notice the Operating system filter chips have an extra close button, where other filters don't. maybe this is another quirk of the `DataViewTreeFilter`? if so, what do you think about opening an issue there to optionally hide it so it's consistent with the other filters?

Image

src/Pages/Repositories/ContentListTable/hooks/useContentListFilters.ts
```diff
@@ -20,7 +17,7 @@ const StatusDisplayMap = {
 
 export const FilterLabelsMap = {
   Search: 'Name/URL',
-  Versions: 'OS version',
+  Versions: 'Operating system',
```
xbhouse, line 20: thank you for updating the OS filter and column names to be consistent with Patch and Inventory! ❤️ could you update these for the templates table too?

src/Pages/Repositories/ContentListTable/ContentListTable.tsx
```diff
@@ -781,13 +787,24 @@ const ContentListTable = () => {
...
                   filterId='versions'
                   ouiaId='filter_version'
-                  aria-label='filter OS version'
                   title={FilterLabelsMap.Versions}
-                  placeholder='Filter by OS version'
-                  options={osFilterOptions}
+                  // TODO: Create an issue in the data-view repository to support a custom aria-label and a placeholder
```
xbhouse, line 794: with the `DataViewTreeFilter`, the filter dropdown closes after any selection instead of staying open until manually closed. (you've mentioned this behavior before, just posting here for transparency 😺). i think this is a valid reason to open a separate issue in react-data-view to also align the close behavior with the checkbox filters

EDIT: ah i see you've opened an issue for this already, thank you!!

xbhouse (commented):
this looks awesome! thank you! ✨ added a couple small comments, and i think any things that would require an update to react-data-view are non-blocking

src/Pages/Repositories/ContentListTable/ContentListTable.tsx
katarinazaprazna: Hi Bryttanie, great catch, thanks for flagging that! It looks like a quirk with this specific filter. It's a button with the name "Close label group Operating system". I’m going to look into whether we can disable it, if not, I’ll get an issue opened for it :)

src/Pages/Repositories/ContentListTable/hooks/useContentListFilters.ts
katarinazaprazna: Yes, will do :)

src/Pages/Repositories/ContentListTable/ContentListTable.tsx
katarinazaprazna: https://github.com/patternfly/react-data-view/issues/644

src/Pages/Repositories/ContentListTable/ContentListTable.tsx
xbhouse: ignore the `any` type comment, restarted vscode and the type is there :sweat_smile:

xbhouse (dismissed):
looks great! thanks for all your work on this :)

xbhouse (approved):

PR #989, opened 2026-05-02
HMS-10608: refine ContentListTable as AI migration model

katarinazaprazna:
## Summary

The `ContentListTable` is currently our only table implemented using data-view. We plan to use AI to migrate the remaining tables to data-view, and this PR polishes the `ContentListTable` so the AI has a clean, well-structured example to work from alongside the general migration guidelines.

- Modularize table components: Extract `RepositoryCell` and `RepositoryActionCell` into their own components, and add a `useRowActions` hook to centralize row action logic. This reduces the main component from ~600 lines and makes the structure easier to follow and replicate.

- Standardize IDs, aria labels, and filters: Adopt kebab-case for all `ouiaId` values, add explicit `aria-label` props, and add `ouiaId` to the toolbar. This is consistent with how data-view components should be wired up. Remove the unused `data-ouia-component-id` prop in favor of the `ouiaId` prop that data-view components consume.

- Align filter placeholder copy: Researched the Console app to align our filter placeholders (e.g. "Filter by name or URL") with the conventions used across other apps.

- Replace custom CSS with PatternFly utilities: Remove custom CSS in favor of properly composing PatternFly UI components and using built-in utility classes.

Screenshot From 2026-05-02 22-57-25

## Notes
I noticed a few tricky spots (things that weren't exactly straightforward) while tweaking the repositories table and testing the AI migration guide on some simpler tables:

- Manual CSS tweaks were still necessary. We have a lot of legacy custom CSS, and the AI usually couldn't figure out how to strip it away on its own. Ideally, we want to drop almost all of that custom styling since DataView is built to handle the design for us.

- TypeScript types were a bit of a struggle. The AI couldn't always type things correctly, so I ended up pointing it toward specific DataView implementation files. This helped it use the actual exported types instead of just making up custom ones.

- Keeping ouiaId unique. To make sure these IDs stay unique across the app, I had to explicitly tell the AI to generate distinct IDs for each table rather than reusing them or sticking with generic names.

- PatternFly guidelines. Sometimes the AI didn't quite get the component composition right. I shared the [PatternFly UI guide](https://github.com/patternfly/patternfly-mcp/tree/main) with it to help keep things aligned.

- For some reason, the AI kept ignoring the instruction to use dynamic imports for DataView components, even though it is in the migration guide. Definitely double check your imports, you might need to remind the AI to stick to the dynamic ones.

- It didn't know how to handle tables with expandable rows because the library doesn't export a specific expandable row component yet. I tried a workaround using a standard PatternFly component, but I want to check with the DataView team to see if they have something official in the works. I will ask for their thoughts on my implementation and see if we can add it to the migration guide. I will keep everyone posted once I hear back.

Next steps: We should think about the best way to reuse repeating logic and UI patterns across our DataView tables. The Inventory team already dealt with this, so I am going to dig into their code to see how they managed it.

## Testing steps

1. Verify the repositories table loads, filters, and pagination work as before
2. Verify bulk actions (delete, introspect) still function correctly       
3. Jest tests pass
4. Playwright tests pass

### DIFF
--- a/src/Pages/Repositories/ContentListTable/hooks/useRowActions.ts
+++ b/src/Pages/Repositories/ContentListTable/hooks/useRowActions.ts
@@ -1,0 +1,167 @@
+import { useCallback } from 'react';
+import { useNavigate } from 'react-router-dom';
+import { IAction } from '@patternfly/react-table';
+import { TooltipPosition } from '@patternfly/react-core';
+import { createUseStyles } from 'react-jss';
+import { ContentOrigin } from 'services/Content/ContentApi';
+import { useAppContext } from 'middleware/AppContext';
+import { EDIT_ROUTE, UPLOAD_ROUTE, DELETE_ROUTE } from 'Routes/constants';
+import type { ActionRowData } from '../components/RepositoryActionCell';
+
+const useStyles = createUseStyles({
+  disabledButton: {
+    pointerEvents: 'auto',
+    cursor: 'default',
+  },
+});
+
+export const showPendingTooltip = (
+  snapshotStatus: string | undefined,
+  introspectStatus: string | undefined,
+) => {
+  if (!snapshotStatus && !introspectStatus) {
+    return 'Introspection or snapshotting is in progress';
+  } else if (snapshotStatus === 'running' || snapshotStatus === 'pending') {
+    return 'Snapshotting is in progress';
+  } else if (introspectStatus === 'Pending') {
+    return 'Introspection is in progress';
+  }
+};
+
+interface UseRowActionsProps {
+  isRedHatRepository: boolean;
+  introspectRepoForUuid: (uuid: string) => Promise<void>;
+  triggerIntrospectionAndSnapshot: (uuid: string) => Promise<void>;
+  clearSelectedRepositories: () => void;
+}
+
+export default function useRowActions({
+  isRedHatRepository,
+  introspectRepoForUuid,
+  triggerIntrospectionAndSnapshot,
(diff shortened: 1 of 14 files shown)


xbhouse:
https://issues.redhat.com/browse/HMS-10608

src/Pages/Repositories/ContentListTable/hooks/useRowActions.ts
```diff
@@ -0,0 +1,155 @@
...
+    (rowData: ActionRowData): IAction[] =>
+      isRedHatRepository ||
+      rowData.origin === ContentOrigin.REDHAT ||
+      rowData.origin === ContentOrigin.COMMUNITY
+        ? features?.snapshots?.accessible
+          ? [
+              {
+                isDisabled: !rowData.snapshot || !(rowData.snapshot && rowData.last_snapshot_uuid),
+
+                ouiaId: 'kebab_view_snapshots',
```
xbhouse: nitpick :) could you also update the ouiaIds for the row actions to be in kebab case?

xbhouse (commented):
thank you! i left one small comment but overall this looks great 😻

src/Pages/Repositories/ContentListTable/hooks/useRowActions.ts
katarinazaprazna: No worries, it definitely helps with consistency. Just pushed the fix now :)

xbhouse (approved):
lgtm!

PR #994, opened 2026-05-05
HMS-9843: Fix template page pagination unhandle error

swadeley:
## Summary
Fix template page pagination unhandle error

Patch template systems returns 400 when offset is past the new total; handle like existing 404 normalization and clamp UI page to meta.
[HMS-9843] [UI] Template page breaks when trying to unassign all systems from a template while being on the last page of the paginated system list

## Testing steps
    Test with UI running in local mode to see the fix
    Register minimum 11 systems to a template in console
    Choose to display 10 systems per page from pagination dropdown
    Select all systems from all the pages to unassign from the template
    Be on the last page and click on Unassign template kebab menu button

[HMS-9843]: https://redhat.atlassian.net/browse/HMS-9843?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Pages/Templates/TemplateDetails/components/Tabs/TemplateSystemsTab.tsx
+++ b/src/Pages/Templates/TemplateDetails/components/Tabs/TemplateSystemsTab.tsx
@@ -55,7 +55,7 @@ const useStyles = createUseStyles({
     justifyContent: 'space-between',
   },
   // Needed to fix styling when "Add repositories" button is disabled
-  ctions: {
+  actions: {
     display: 'flex',
     flexDirection: 'row',
   },
@@ -224,7 +224,7 @@ export default function TemplateSystemsTab() {
                 customIcon={<SearchIcon />}
               />
             </InputGroupItem>
-            <FlexItem className={classes.ctions}>
+            <FlexItem className={classes.actions}>
               <ConditionalTooltip
                 content={`You do not have the required ${missingRequirements} to perform this action.`}
                 show={isMissingRequirements}
--- a/src/Pages/Templates/TemplateDetails/components/Tabs/TemplateSystemsTab.tsx
+++ b/src/Pages/Templates/TemplateDetails/components/Tabs/TemplateSystemsTab.tsx
@@ -63,6 +63,21 @@ const useStyles = createUseStyles({
 
 const perPageKey = 'TemplateSystemsPerPage';
 
+export const calculatePageAfterSystemsDelete = (
+  currentPage: number,
+  perPage: number,
+  totalItems: number,
+  deletedItemsCount: number,
+) => {
+  if (deletedItemsCount <= 0) {
+    return currentPage;
+  }
+
+  const nextTotalItems = Math.max(0, totalItems - deletedItemsCount);
+  const maxPage = Math.max(1, Math.ceil(nextTotalItems / perPage));
+  return Math.min(currentPage, maxPage);
+};
+
 export default function TemplateSystemsTab() {
   const classes = useStyles();
   const queryClient = useQueryClient();
@@ -130,13 +145,6 @@ export default function TemplateSystemsTab() {
     meta: { total_items = 0 },
   } = data;
 
-  useEffect(() => {
-    const maxPage = Math.max(1, Math.ceil(total_items / perPage));
-    if (page > maxPage) {
-      setPage(maxPage);
-    }
-  }, [page, perPage, setPage, total_items]);
-
   const allSelected = useMemo(
     () => systemsList.every(({ inventory_id }) => selectedList.has(inventory_id)),
     [selectedList, systemsList],
@@ -167,8 +175,16 @@ export default function TemplateSystemsTab() {
     }
   };
 
-  const deselectAll = () => {
-    setSelected([]);
+  const handleDeleteSystems = async (items: string[]) => {
+    const nextPage = calculatePageAfterSystemsDelete(page, perPage, total_items, items.length);
+    const deletedSystems = new Set(items);
+
+    if (nextPage !== page) {
+      setPage(nextPage);
+    }
+
+    await deleteFromSystems(items);
+    setSelected((prev) => prev.filter((selectedId) => !deletedSystems.has(selectedId)));
   };
 
   const onPerPageSelect = (_, newPerPage, newPage) => {
@@ -253,8 +269,7 @@ export default function TemplateSystemsTab() {
               setDisabled
             >
               <SystemsDeleteKebab
-                deleteFromSystems={deleteFromSystems}
-                deselectAll={deselectAll}
+                deleteFromSystems={handleDeleteSystems}
                 isDisabled={!rbac?.templateWrite || selected.length > TEMPLATE_SYSTEMS_UPDATE_LIMIT}
                 selected={selected}
               />
@@ -337,7 +352,7 @@ export default function TemplateSystemsTab() {
           allSelected={allSelected}
           perPage={perPage}
           selectAllToggle={selectAllToggle}
-          deleteFromSystems={deleteFromSystems}
+          deleteFromSystems={handleDeleteSystems}
           isLoadingOrDeleting={loadingOrDeleting}
           systemsList={systemsList}
           sortParams={sortParams}
--- a/src/Pages/Templates/TemplateDetails/components/Tabs/TemplateSystemsTab.tsx
+++ b/src/Pages/Templates/TemplateDetails/components/Tabs/TemplateSystemsTab.tsx
@@ -108,13 +108,17 @@ export default function TemplateSystemsTab() {
     [activeSortIndex, activeSortDirection],
   );
 
-  const { hasCompatibleSystems, isFetchingCompatibility, isCompatibilityError } =
-    useCompatibleSystems(uuid);
+  const {
+    hasCompatibleSystems,
+    isFetchingCompatibility,
+    isCompatibilityError,
+    compatibilityError,
+  } = useCompatibleSystems(uuid);
 
   const {
-    isLoading,
-    error,
-    isError,
+    isTemplateSystemsLoading,
+    templateSystemsError,
+    isTemplateSystemsError,
     data = { data: [], meta: { total_items: 0, limit: 20, offset: 0 } },
   } = useListSystemsByTemplateId(uuid, page, perPage, debouncedSearchQuery, sortString);
 
@@ -125,6 +129,13 @@ export default function TemplateSystemsTab() {
     data: systemsList = [],
     meta: { total_items = 0 },
   } = data;
+
+  useEffect(() => {
+    const maxPage = Math.max(1, Math.ceil(total_items / perPage));
+    if (page > maxPage) {
+      setPage(maxPage);
+    }
+  }, [page, perPage, setPage, total_items]);
 
   const allSelected = useMemo(
     () => systemsList.every(({ inventory_id }) => selectedList.has(inventory_id)),
@@ -166,9 +177,7 @@ export default function TemplateSystemsTab() {
     localStorage.setItem(perPageKey, newPerPage.toString());
   };
 
-  const loadingOrDeleting = isLoading || isDeleting;
-
-  const errorState = isError || isCompatibilityError;
+  const loadingOrDeleting = isTemplateSystemsLoading || isDeleting;
 
   const sortParams = (columnIndex: number): ThProps['sort'] | undefined =>
     columnSortAttributes[columnIndex]
@@ -191,12 +200,12 @@ export default function TemplateSystemsTab() {
   const missingRequirements =
     rbac?.templateWrite && !hasRHELSubscription ? 'subscription (RHEL)' : 'permission';
 
-  if (isLoading || isFetchingCompatibility) {
+  if (isTemplateSystemsLoading || isFetchingCompatibility) {
     return <Loader />;
   }
 
-  if (errorState) {
-    throw error;
+  if (isTemplateSystemsError || isCompatibilityError) {
+    throw templateSystemsError ?? compatibilityError;
   }
 
   return (
@@ -286,7 +295,7 @@ export default function TemplateSystemsTab() {
                       id='addSystemsButton'
                       ouiaId='add_systems'
                       variant='primary'
-                      isDisabled={isLoading}
+                      isDisabled={isTemplateSystemsLoading}
                       onClick={() => {
                         const method = hasCompatibleSystems
                           ? '' // If there are some registered systems, open the system list view
--- a/src/services/Systems/SystemsApi.ts
+++ b/src/services/Systems/SystemsApi.ts
@@ -151,33 +151,6 @@ const emptyTemplateSystemsResponse = (limit: number, offset = 0): IDSystemsColle
   },
 });
 
-/** Patch 400 when offset is past the list (JSON:API errors or legacy `{ error: string }`). */
-const isInvalidOffsetPatchAxiosError = (err: unknown): boolean => {
-  if (!axios.isAxiosError(err) || err.response?.status !== 400) {
-    return false;
-  }
-
-  const data = err.response.data as
-    | {
-        error?: string;
-        errors?: Array<{ code?: string; source?: { pointer?: string; parameter?: string } }>;
-      }
-    | undefined;
-
-  const firstError = data?.errors?.[0];
-  // Backend currently returns { error: "Invalid offset" } for this case.
-  const isLegacyInvalidOffsetMessage = data?.error === 'Invalid offset';
-  const isInvalidOffsetCode = firstError?.code === 'INVALID_OFFSET';
-  const isOffsetParameter =
-    firstError?.source?.parameter === 'offset' || firstError?.source?.pointer === '/meta/offset';
-
-  return Boolean(isLegacyInvalidOffsetMessage || isInvalidOffsetCode || isOffsetParameter);
-};
-
-/** Invalid offset on page > 1 after the list shrinks (e.g. bulk unassign). */
-const isRecoverableTemplateSystemsPaginationError = (err: unknown, page: number): boolean =>
-  page > 1 && isInvalidOffsetPatchAxiosError(err);
-
 export const getSystemsList: (
   page: number,
   limit: number,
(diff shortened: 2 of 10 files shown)


xbhouse:
https://issues.redhat.com/browse/HMS-9843

swadeley:
Hi

setting  to draft status while I investigate API updates in patchman-engine

swadeley:
@sourcery-ai review

swadeley:
@sourcery-ai review

swadeley:
@sourcery-ai review

swadeley:
@sourcery-ai review

swadeley:
@sourcery-ai review

swadeley:
@sourcery-ai review

swadeley:
@sourcery-ai review

swadeley:
@sourcery-ai review

swadeley:
@sourcery-ai review

swadeley:
@sourcery-ai review

src/Pages/Templates/TemplateDetails/components/Tabs/TemplateSystemsTab.tsx
```diff
@@ -126,6 +130,13 @@ export default function TemplateSystemsTab() {
     meta: { total_items = 0 },
   } = data;
 
+  useEffect(() => {
+    const maxPage = Math.max(1, Math.ceil(total_items / perPage) || 1);
```
Dugowitch: Hey, what is the purpose of  the ` || 1` part? Maybe I'm missing something, but it seems redundant. It will only kick in when the left side is zero, but `Math.max(1, 0)` already handles that case, or when the left side is `NaN`, which should not happen

swadeley: OK, removing now, thank you

src/Pages/Templates/TemplateDetails/components/Tabs/TemplateSystemsTab.tsx
```diff
@@ -166,9 +177,9 @@ export default function TemplateSystemsTab() {
     localStorage.setItem(perPageKey, newPerPage.toString());
   };
 
-  const loadingOrDeleting = isLoading || isDeleting;
+  const loadingOrDeleting = isTemplateSystemsLoading || isDeleting;
 
-  const errorState = isError || isCompatibilityError;
+  const errorState = isTemplateSystemsError || isCompatibilityError;
```
Dugowitch: Just wondering if we need `errorState` const if it's only used once

src/Pages/Templates/TemplateDetails/components/Tabs/TemplateSystemsTab.tsx
```diff
@@ -191,12 +202,17 @@ export default function TemplateSystemsTab() {
...
 
   if (errorState) {
-    throw error;
+    if (isTemplateSystemsError) {
+      throw templateSystemsError ?? new Error('Unable to fetch systems assigned to this template');
+    }
+    if (isCompatibilityError) {
+      throw compatibilityError ?? new Error('Unable to check compatible systems for this template');
+    }
   }
```
Dugowitch: The `errorState` check actually became redundant with the addition of specific isFooError checks inside, since they cover all of the options.

Also, the description for fooError is `The error object for the query, if an error was thrown.` - I think it's safe to assume the fooError is set if isFooError is true.

I'd write the if clause concisely like this:
```ts
if (isTemplateSystemsError || isCompatibilityError) {
  throw templateSystemsError ?? compatibilityError
}
```
but having two simple if clauses might be more readable.

src/services/Systems/SystemsQueries.ts
```diff
@@ -11,8 +12,29 @@ import {
...
+ */
+function fingerprintInventoryIds(ids: readonly string[] | undefined): string {
+  if (!ids?.length) return '';
+  const sorted = [...ids].sort();
+  let h = 0;
+  for (const id of sorted) {
+    for (let i = 0; i < id.length; i++) {
+      h = (Math.imul(31, h) + id.charCodeAt(i)) | 0;
+    }
+  }
```
Dugowitch: Since the whole first and last items are in the final key, we could omit them from the hash calculation by starting from `1` and going to `length - 1`, slightly reducing the number of iterations and slightly improving the information density of the hash

src/services/Systems/SystemsQueries.ts
```diff
@@ -11,8 +12,29 @@ import {
...
+  const first = sorted[0] ?? '';
+  const last = sorted[sorted.length - 1] ?? '';
+  return `${sorted.length}:${(h >>> 0).toString(16)}:${first}:${last}`;
```
Dugowitch: The `?? ''` seems redundant, since at the beginning of the function, there is a check that the array is not empty

Dugowitch: I believe it'd be sufficient to use just `h` instead of `(h >>> 0)` - it should be an OK number

src/services/Systems/SystemsApi.ts
```diff
@@ -139,79 +149,251 @@ export interface PatchTemplatesResponse {
...
+  if (Array.isArray(body.data) && body.data.length > 0) {
+    return body.data
+      .map((row) => row.id ?? row.inventory_id ?? '')
+      .filter((id): id is string => id.length > 0);
+  }
+  if (Array.isArray(body.ids) && body.ids.length > 0) {
+    return body.ids.filter((id) => id.length > 0);
+  }
+  return [];
+};
```
Dugowitch: As ~~meantioned before~~ stated in the next comment, PatchTemplateSystemsIdsResponse will have both `data` and `ids`, and they will contain the same ids. So in this case, it will be easier to work with just `ids`. Also, I do not think an id would be empty string, since those are DB primary keys (or part thereof)

I'd write it like this:
```ts
return body.ids ?? []
```
in which case, one might argue the function is not even necessary.

src/services/Systems/SystemsApi.ts
```diff
@@ -87,6 +87,16 @@ export interface SystemsCollectionResponse {
...
 
+/**
+ * Response from GET `/ids/templates/{id}/systems`.
+ * Patch returns JSON:API `data` resources with `id` (see test-utils `resolvePatchSystemIdsForHostname`);
+ * some responses may instead expose a plain `ids` array.
+ */
+export interface PatchTemplateSystemsIdsResponse {
+  ids?: string[];
+  data?: Array<{ id?: string; inventory_id?: string }>;
+}
```
Dugowitch: Patch will actually always return both with the same ids, just structured differently, see https://github.com/RedHatInsights/patchman-engine/blob/cf0901c2b3b64f6aae0280121dc65cc68e656676/manager/controllers/utils.go#L509-L514

Also, the object in `data` array does not have `inventory_id` key, see https://github.com/RedHatInsights/patchman-engine/blob/cf0901c2b3b64f6aae0280121dc65cc68e656676/manager/controllers/structures.go#L36-L38
Response to `/templates/{template_id}/systems` has it, but one to `/ids/...` does not.

src/services/Systems/SystemsApi.ts
```diff
@@ -139,79 +149,251 @@ export interface PatchTemplatesResponse {
...
+      const maxPage = Math.max(1, Math.ceil(total / limit));
+      const targetPage = Math.min(page, maxPage);
+      if (targetPage <= 1) {
+        return firstPage;
+      }
+      return load(targetPage);
     }
     throw err;
   }
 };
```
Dugowitch: I have trouble understanding why switch from querying `/template/{templateID}/systems` to querying `/systems` with a filter for all systemIDs that are in the template

src/services/Systems/SystemsApi.ts
```diff
@@ -139,79 +149,251 @@ export interface PatchTemplatesResponse {
...
+        return offset === 0 ? [] : all;
+      }
+      // List can shrink while paging (e.g. unassign); invalid offset on a later chunk is end-of-list.
+      if (offset > 0 && isInvalidOffsetPatchAxiosError(err)) {
+        return all;
+      }
+      throw err;
+    }
+  }
+};
```
Dugowitch: Do I understand correctly that this function loads all system ids for a template iteratively page after page?

Dugowitch (changes requested):

src/services/Systems/SystemsApi.ts
swadeley: Hi, 
the main goal[1] was to switch from GET to POST[2] and "instead of requesting systems as filter[id]=in:uuid1,uuid2,… put them into request body as json: {“ids”:[ “uuid1”, “uuid2”, …]}".

From the RHINENG-23971 PR I referred to[2]:
"In patchman-engine, POST /systems is documented as:

Same response as GET /systems, restricted to the given inventory IDs. Use query parameters for paging, sort, search, and filters."

So I am doing what I think the comments in the Jira[1] are asking me to do.

[1]https://redhat.atlassian.net/browse/HMS-9843?focusedCommentId=16905898
[2] https://github.com/RedHatInsights/patchman-engine/pull/2183/changes

src/Pages/Templates/TemplateDetails/components/Tabs/TemplateSystemsTab.tsx
swadeley: Hi, yes, errorState was only used in that one `if`, so it did not add much. Removed. Thank you.

src/Pages/Templates/TemplateDetails/components/Tabs/TemplateSystemsTab.tsx
swadeley: OK, I used your example clause. Thank you.

src/services/Systems/SystemsQueries.ts
swadeley: Oh, I see, thank you for pointing that out.

src/services/Systems/SystemsQueries.ts
swadeley: OK, sorted always has at least one element, fixed. Thank you.

swadeley: Updated to `h.toString(16)`

thank you.

src/services/Systems/SystemsApi.ts
swadeley: OK, so:

Removed `inventoryIdsFromTemplateSystemsIdsBody` and inlined `const batch = data.ids ?? []` in `fetchAllTemplateSystemsInventoryIds`.

Simplified `PatchTemplateSystemsIdsResponse` to only `ids?: string[]` (no data branch).

src/services/Systems/SystemsApi.ts
swadeley: Hi, simplified PatchTemplateSystemsIdsResponse, please review that again.

src/services/Systems/SystemsApi.ts
swadeley: Yes. Page through /ids/templates/.../systems until we have every inventory UUID, then use that list in the POST /systems body for the template systems tab.  So this approach followed from the use of POST /systems; if that approach is wrong or not the best, then this can change.

Dugowitch: I'm worried about performance impact of this approach, I will check with Mim - what did he mean by the Jira comment you mentioned above

src/services/Systems/SystemsApi.ts
```diff
@@ -139,79 +152,236 @@ export interface PatchTemplatesResponse {
...
+      const maxPage = Math.max(1, Math.ceil(total / limit));
+      const targetPage = Math.min(page, maxPage);
+      if (targetPage <= 1) {
+        return firstPage;
+      }
+      return load(targetPage);
     }
     throw err;
   }
 };
```
Dugowitch: But this endpoint does not use the filter, it does not need to be updated to the POST endpoint

src/services/Systems/SystemsApi.ts
```diff
@@ -139,79 +152,236 @@ export interface PatchTemplatesResponse {
...
+      { ids: filter.ids },
+    );
+    return data;
+  }
+
+  const { data } = await axios.get<SystemsCollectionResponse>(
+    `${patchApiVersionUrl}/systems?${queryString}`,
   );
   return data;
 };
```
Dugowitch, line 189: The `/systems` query here does use the filter, so it is correctly updated to the POST endpoint

src/services/Systems/SystemsQueries.ts
```diff
@@ -52,21 +75,56 @@ export const useListSystemsByTemplateId = (
...
     staleTime: 25_000,
     refetchOnWindowFocus: 'always',
     refetchOnMount: 'always',
     refetchInterval: 20_000,
     meta: {
       title: 'Unable to fetch systems assigned to this template',
-      id: 'systems-list-error',
+      id: 'template-systems-page-query-error',
     },
   });
```
Dugowitch, line 68: and this query does not use the filters, so it does not need to be updated either

Dugowitch (changes requested):
I checked the comment and with Mim and I think there has been a misunderstanding. The change mentioned in the comment is, in fact. only about preventing the GET url from getting too long, because the `filter[id]=in:<list of ids>` part sometimes listed so many ids that the url length caused 400 errors. These may be some of the 400s from HMS-9843 but it's not the main issue. The main cause, if I understand correctly, is that after the systems get unassigned (unassigning works as you mentioned in [this comment](https://redhat.atlassian.net/browse/HMS-9843?focusedCommentId=16767485) \[1\]), there is an error caused by fetching/reloading a page that is no longer valid (because the systems were unassigned). 

\[1\] https://redhat.atlassian.net/browse/HMS-9843?focusedCommentId=16767485

src/services/Systems/SystemsApi.ts
swadeley: Ok, so, POST only where filter[id] is used.

Thank you for the clarification.

swadeley: Change made, thank you.

src/services/Systems/SystemsApi.ts
swadeley: OK, thank you.

src/services/Systems/SystemsQueries.ts
swadeley: Change made, thank you.

src/Pages/Templates/TemplateDetails/components/Tabs/TemplateSystemsTab.tsx
```diff
@@ -126,6 +130,13 @@ export default function TemplateSystemsTab() {
     meta: { total_items = 0 },
   } = data;
 
+  useEffect(() => {
+    const maxPage = Math.max(1, Math.ceil(total_items / perPage));
+    if (page > maxPage) {
+      setPage(maxPage);
+    }
+  }, [page, perPage, setPage, total_items]);
+
```
Dugowitch: Does this block actually achieve something?

src/services/Systems/SystemsApi.ts
```diff
@@ -139,79 +139,128 @@ export interface PatchTemplatesResponse {
...
-      };
+    // After bulk unassign, the UI can still be on a page whose offset is past the new total
+    if (isRecoverableTemplateSystemsPaginationError(err, page)) {
+      const firstPage = await requestPage(1);
+      const total = firstPage.meta?.total_items ?? 0;
+      const maxPage = Math.max(1, Math.ceil(total / limit));
+      const targetPage = Math.min(page, maxPage);
+      if (targetPage <= 1) {
+        return firstPage;
+      }
```
Dugowitch: The whole sequence of requests is not that cheap, I'd still be somewhat worried about performance here. Having thought about the problem, I came up with something:

`TemplateSystemsTab` component is the place where deletion callback is passed on (lines 247 and 331 pass on `deleteFromSystems`) and is the place with all the data - selected systems, their count, page size, current page tracking, total_items, ... - to determine if stay on the page, go to the previous one, or to the first one as a fallback. So are these changes solving the issue in the right place?

I'd propose implementing something like `handleDeleteSystems` in `TemplateSystemsTab.tsx` that would encapsulate the decision logic and call `deleteFromSystems` as well as `setPage` with the calculated page. `handleDeleteSystems` would be passed on to `SystemDeleteKebab` and `SystemsTable` instead of `deleteFromSystems`.

Examples:
- there are 2 pages, current page is 2 with 1 system, on unassign of the system -> load the previous page instead
- current page is = 4, page_size = 20, total_items = 90, selected count = 30, on bulk unassign load page ceil( (90-30) / 20 ) = 3.

When I think about it more, the heuristic might not be that simple, because the systems can be selected on previous and following pages, and users will expect some continuity. A safe option might be *if selected_count >= page_size, load page 1*. I'm sure you can come up with something better than me :wink: However, what is definitely possible is calculating a valid page.

I'm definitely not saying this is THE way to go, just thinking about alternatives. 
What do you think about the idea? Have you thought about resolving the issue there? Would it work?

Dugowitch (changes requested):

src/Pages/Templates/TemplateDetails/components/Tabs/TemplateSystemsTab.tsx
swadeley: Hi,  
The purpose is to set the user to the last valid page using `setPage(maxPage)`, if the values on the last line change.
React watches `[page, perPage, setPage, total_items]` and the effect re-runs whenever any of them changes. But setPage should not change outside of this function so I think the lint checkers are requiring it. I will remove it and test.

Hope I understood you question correctly?

The useEffect alone does not fix the bug, right after delete, the request fails before `total_items` updates. The API recovery prevents that failed refetch from surfacing as an error. Once on page one the useEffect will work to get us to the last valid page.

src/services/Systems/SystemsApi.ts
swadeley: Hi, I would have to test that in another branch.

swadeley: Hi @Dugowitch , changes made. 
I made them as separate commits to help you review the new changes.  Will squash on merge. 
Thank you.

P.S.: code changes reviewed locally with sourcery-ai

src/Pages/Templates/TemplateDetails/components/Tabs/TemplateSystemsTab.tsx
Dugowitch: Sorry, my comment was a little vague. I think the useEffect might be obsolete:
- change of `page` and `perPage` worked before (handled for example by `onPerPageSelect`)
- change of `total_items` rerenders the component anyway - delete call also invalidates `GET_TEMPLATE_SYSTEMS_KEY` so React Query refetches `useListSystemsByTemplateId` and there will be the new value of `total_items`

I removed it and tested manually with 11 items total and 10 per page: went to the 2nd page -> unassigned the system -> was automatically redirected to the first page as expected

swadeley: OK, I have removed that. Thank you.

src/services/Systems/SystemsApi.ts
```diff
@@ -139,79 +139,128 @@ export interface PatchTemplatesResponse {
...
+      return requestPage(targetPage);
     }
```
Dugowitch: This error is prevented by the new changes. I reverted `listSystemsByTemplateId` locally and did a few manual tests and things seemed to work fine. Can it be removed?

Dugowitch (changes requested):
The new changes look good! Just one more question
_(Hope you don't hate me by now 😅)_

src/services/Systems/SystemsApi.ts
swadeley: Hi , `handleDeleteSystems` runs before the delete and refetch, so `listSystemsByTemplateId` should never be called with an out-of-range page after systems being unassigned.  I'll remove it.

Dugowitch (approved):

PR #996, opened 2026-05-06
HMS-10608: prep shared hooks and components for data-view migration

katarinazaprazna:
## Summary

- Restructured advisories table internals to prepare for data-view migration
- Extracted reusable table hooks into Hooks/tables and adopted them in all tables already converted to data-view

Screenshot From 2026-07-07 18-02-14

## Notes

The full advisories table migration is blocked on two upstream issues:

- **Expandable rows**: data-view currently only supports compound expandable rows. Our tables use traditional single-expand rows, so migrating now requires workarounds
- **Stale states on page/filter/sort changes**: the basic table tracks rows by index rather than by ID, so when the user pages, sorts, or filters, the row at the same index retains the previous row's expanded/styled state

Once both issues are resolved upstream, we can migrate the advisories table cleanly without workarounds. This PR lands the prep work so the migration itself will be a small, focused change

## Testing steps

- Pagination works across all data-view tables (repositories, packages modal, snapshot packages, template packages, advisories)
- Loading and empty states display as expected


xbhouse:
https://issues.redhat.com/browse/HMS-10608

src/components/SharedTables/AdvisoriesTable/components/RebootStatus.tsx
```diff
@@ -0,0 +1,30 @@
...
+import text from '@patternfly/react-styles/css/utilities/Text/text';
+
+interface Props {
+  rebootSuggested: boolean;
+}
+
+export default function RebootStatus({ rebootSuggested }: Props) {
+  const color = rebootSuggested
+    ? t_global_color_status_danger_100.value
+    : t_global_color_disabled_100.value;
```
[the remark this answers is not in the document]
katarinazaprazna: This is being investigated with UX as I found conflicting PF guidelines about this

src/components/SharedTables/AdvisoriesTable/index.tsx
```diff
@@ -1,226 +1,234 @@
...
+                <>
+                  <AngleDownIcon aria-hidden className={classes.expandToggleIcon} />
+                  <span className='pf-v6-screen-reader'>Details</span>
+                </>
+              ),
+            } as DataViewTd & { id: number },
+            { cell: errata_id },
+            { cell: reduceStringToCharsWithEllipsis(summary, 70) },
+            { cell: <ErrataTypeCell type={type} /> },
+            { cell: <SeverityCell severity={severity} /> },
```
xbhouse: it looks like filtering the advisory table in some cases displays stale severity icon colors based on the row index. this might have been pre-existing, but the old table used `errata_id` in the row keys so maybe it never showed up? one way to fix this could be by adding `errata_id` as a key on SeverityCell to force it to remount. what do you think?

src/components/SharedTables/AdvisoriesTable/index.tsx
```diff
@@ -1,226 +1,234 @@
...
+            { cell: formatDateDDMMMYYYY(issued_date) },
+          ],
+        });
 
-  useDeepCompareEffect(() => {
-    if (!isEmpty(expandState)) setExpandState({});
-  }, [errataList]);
+        expandableContents.push({
```
xbhouse: if a row is expanded and the user changes pages or filters or sorts, the row at the same index on the new state is expanded. this seems like it's also related to DataView using the row index to track state changes, maybe the expandable content could be reset or remounted with a key here too?

katarinazaprazna: Thanks for calling this out! The stale states aren't something we can cleanly fix on our end. The basic table tracks rows by index rather than ID (unlike the tree table variant), which causes both the stale expansion and severity colours on filter/sort/page changes. I'm going to sync with the data-view team to see when they plan to adopt ID-based tracking in the basic table

We could add a dynamic key prop to the table to force a full remount on every filter/sort/page update, but that feels little hacky. It might be better to exclude the Advisories table from this PR until this is resolved. What do you think?

katarinazaprazna: Hi @xbhouse, all set! I created a GitHub [issue](https://github.com/patternfly/react-data-view/issues/677) for that table behaviour :)

xbhouse: thanks @katarinazaprazna ! i agree it makes sense to fix this on the data view side, especially if the desired behavior already exists in other table types. because of that and the other expandable row workarounds, i do think it's better to wait on migrating the advisories table

src/components/SharedTables/AdvisoriesTable/components/RebootStatus.tsx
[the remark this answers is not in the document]
katarinazaprazna: UX Update: Let's use the `--pf-t--global--icon--color--regular` token in the `RebootStatus` component. The Remediations app is currently using grey (disabled), which is incorrect. UX will follow up with the Remediations team to get this fixed @marisvirik

Starle21 (commented):
Great work in the direction of making our code more reusable! 🎉 

From the collocation perspective, since the components (`ErrataTypeCell`, `RebootStatus`, `SeverityCell`) are used only in the Advisories table, it is a good move to nest them under it!

I have a few points for possible improvement:

## 1. usage of `Tables` v. `Hooks/tables`

The new Tables folder was created with the goal to be the place to host all the new DataView tables eventually. All the things that the tables need to function - hooks, constants and such - is meant to be collocated there. 
Since a table hook needs a table to function and cannot be used separately, it does not make sense to place it under Hooks. Hooks folder signifies reusability over possibly any component, not just table components, AFAIK.

**So I suggest to put the two hooks placed in `Hooks/tables` to `Tables/Generic/hooks`.**

## 2. usage of `Tables` v. `SharedTables`

SharedTables is the legacy folder for non-dataview tables. It should go away eventually as we refactor tables to dataview tables. So I don't think it makes sense to extend it.

**So I suggest to put the new code for dataview stuff under `Tables/Advisories` (useAdvisoriesTable hook, constants, components).**

Note:
_The aim we have talked previously about was also about generalizing tables, so that they are consistent and reusable in multiple places._

_That is a big goal right now, since there are many different table types used. As a first step, we have agreed it would not make sense to generalize everything. But start generalizing along the different types of tables for `Packages`, `Advisories`, `Systems`, `Repositories`, etc._

_Collocating the different types of tables first will help us see the same shared patterns and will help us extract reusable code even more._

**_So from this perspective let's put Advisories under Tables. And in the future also other table types._**

## 3. Naming - folders and files under Tables

It is good to keep names as short as possible, to reduce the fatigue or reading too long names, without loosing meaning though. 

Placing things under Tables folder signifies that everything under it has something to do with Tables, so there is no point to repeat "Tables" in every folder or file name.

**So I agree it was a good call to remove "Tables" from `useTablePaginationLocalStorage`.
In the same way I suggest to remove "Table" from `useTableActiveState`.**

I would keep the 'Table' name only for the components that actually have the tables inside of them.

## 4. Better types definition for ErrataTypeCell

Right now ErrataTypeCell type says that it accepts any kind of string for the type argument. But this is not so. It expects only specific string literals, so it can choose a proper icon by the literal, otherwise it shows UnknownIcon.

**I suggest therefore to use the specific string literals for the `type` argument.** 

It will also help the Typescript static code analysis. So in the end there is no need to have Unknown type defined, since the type checker would flag it if a not defined icon type was being passed in.

## 5. constants uppercase

**columnsConfig is a defined constant, so let's use all uppercase for that.**

katarinazaprazna:
Hi @Starle21, thanks for the feedback!

Here is my thought process on a couple of these points:

**Tables vs. Hooks/tables:** The idea here was to reserve Hooks/tables strictly for generic, reusable hooks that any table could consume. For component-specific hooks, I like keeping them co-located with the table itself. For example: `AdvisoriesTable/useAdvisoriesTable.ts` right next to `AdvisoriesTable/index.tsx`. How do you feel about that?

**Hook Naming:** I intentionally kept "table" in `useTableActiveState` because `useActiveState` felt a bit too generic on its own (whereas something like `usePaginationLocalStorage` is already pretty explicit). Keeping "table" in the name makes its purpose immediately clear at a glance, but I'm open to ideas

Just to give a bit more context, this PR is mostly foundational prep work. Once the two upstream PRs land, we'll be able to properly implement the actual Advisories table, including expandable rows. To keep this PR manageable, would you be open to addressing these refinements in that follow-up PR where we tackle the full implementation?

Starle21:
> **Tables vs. Hooks/tables:** The idea here was to reserve Hooks/tables strictly for generic, reusable hooks that any table could consume. For component-specific hooks, I like keeping them co-located with the table itself. For example: `AdvisoriesTable/useAdvisoriesTable.ts` right next to `AdvisoriesTable/index.tsx`. How do you feel about that?

I think it makes sense to colocate it next to each other. I would put it into Table/Advisories/hooks, but I do not have a strong preference :)

> **Hook Naming:** I intentionally kept "table" in `useTableActiveState` because `useActiveState` felt a bit too generic on its own (whereas something like `usePaginationLocalStorage` is already pretty explicit). Keeping "table" in the name makes its purpose immediately clear at a glance, but I'm open to ideas

Yeah, I think it is about preference or. I think that it is understandable from the import, that it is related to the table, but that might involve a scroll up or going into the implementation. I would keep the name short.

> Just to give a bit more context, this PR is mostly foundational prep work. Once the two upstream PRs land, we'll be able to properly implement the actual Advisories table, including expandable rows. To keep this PR manageable, would you be open to addressing these refinements in that follow-up PR where we tackle the full implementation?

Sure, let's do that there :)

Starle21 (approved):

PR #1004, opened 2026-05-12
Test: Set dependabot ignore for major updates to ts-patch

swadeley:
## Summary
Set dependabot ignore for major updates to ts-patch
To prevent conflict with Typescript version


xbhouse (approved):
thanks!

PR #1006, opened 2026-05-12
HMS-9614: UI for remove rpm packages

Starle21:
## Summary
Users can now delete rpm packages in the Package Modal
accessible from the repositories list. Only upload
repositories can be deleted.

The underlying packages table was changed to use DataView
components, so bulk delete can be enabled. Pagination
and filter by name was also delegated to DataView,
which made the consuming components significantly shorter.
Delete bulk and delete single package is only visible
for Package Modal.

This is the UI for the backend endpoint `/api/content-sources/v1.0/repositories/${repoUuid}/rpms/bulk_remove/` that was recently added.

## Testing steps
1) All the tests pass, specifically `PackageModal.test.tsx` unit test and UI tests in `DeletePackages.spec.ts`.
2) Manually go and create upload repo, add packages there and then go and delete a package or a couple of packages. Confirm that everything works.

Note: AssignEUSTemplateToSystem is failing currently due to an unrelated reason.

#testwith https://github.com/content-services/content-sources-backend/pull/1499

### DIFF
--- a/_playwright-tests/UI/DeletePackages.spec.ts
+++ b/_playwright-tests/UI/DeletePackages.spec.ts
@@ -1,0 +1,170 @@
+import path from 'path';
+import type { Page } from '@playwright/test';
+
+import {
+  test,
+  expect,
+  cleanupRepositories,
+  randomName,
+  waitWhileRepositoryIsPending,
+} from 'test-utils';
+import { BULK_TASK_TIMEOUT_MS } from '../testConstants';
+import {
+  closeGenericPopupsIfExist,
+  getRowByNameOrUrl,
+  retry,
+  waitForValidStatus,
+} from './helpers/helpers';
+import { navigateToRepositories } from './helpers/navHelpers';
+
+const uploadRepoNamePrefix = 'package-modal-delete-upload';
+const externalRepoNamePrefix = 'package-modal-delete-external';
+const externalRepoUrl = 'https://jlsherrill.fedorapeople.org/fake-repos/signed/';
+const packageName = 'bear';
+
+const openPackagesModal = async (page: Page, repoName: string) => {
+  const row = await getRowByNameOrUrl(page, repoName);
+  await row.getByTestId('package_count_button').click();
+  const packagesModal = page.getByRole('dialog', { name: 'Packages' });
+  await expect(packagesModal).toBeVisible();
+  await expect(packagesModal.getByTestId('packages_table')).toBeVisible();
+  return packagesModal;
+};
+
+test.describe('PackageModal delete packages', () => {
+  test('Upload repository can delete a package from PackageModal', async ({
+    page,
+    client,
+    cleanup,
+  }) => {
+    const uploadRepoName = `${uploadRepoNamePrefix}-${randomName()}`;
+
+    await cleanup.runAndAdd(() => cleanupRepositories(client, uploadRepoNamePrefix));
+    await closeGenericPopupsIfExist(page);
+    await navigateToRepositories(page);
+
+    await test.step('Create upload repository and upload an RPM', async () => {
+      // click add button
+      await page.getByRole('button', { name: 'Add repositories' }).first().click();
+
+      // fill name and check upload checkbox, that is all that is needed
+      await page.getByPlaceholder('Enter name').fill(uploadRepoName);
+      await page.getByLabel('Upload', { exact: true }).check();
+
+      // click save and other modal will get shown
+      const [, bulkCreateResponse] = await Promise.all([
+        page.getByRole('button', { name: 'Save and upload content' }).click(),
+        page.waitForResponse(
+          (resp) =>
+            resp.url().includes('/bulk_create/') && resp.status() >= 200 && resp.status() < 300,
+          { timeout: BULK_TASK_TIMEOUT_MS },
+        ),
+      ]);
+
+      // wait until check repository status becomes valid and "drag and drop" with upload button shows
+      const bulkCreateData = await bulkCreateResponse.json();
+      const repoUuid = bulkCreateData[0]?.uuid;
+      expect(repoUuid).toBeTruthy();
+      const repo = await waitWhileRepositoryIsPending(client, repoUuid);
+      expect(repo.status).toBe('Valid');
+      await expect(page.getByText('Drag and drop files here')).toBeVisible();
+
+      // put in the file to upload
+      const filePath = path.join(__dirname, './fixtures/bear-4.1-1.noarch.rpm');
+      await retry(page, async (page) => {
+        await page.locator('input[type=file]').first().setInputFiles(filePath);
+      });
+      await expect(page.getByText('All uploads completed!')).toBeVisible();
+      await page.getByRole('button', { name: 'Confirm changes' }).click();
+
+      await waitForValidStatus(page, uploadRepoName);
+    });
+
+    await test.step('Delete package via row actions in PackageModal', async () => {
+      // click on number of package to open the packages modal
+      let packagesModal = await openPackagesModal(page, uploadRepoName);
+
+      let row = await getRowByNameOrUrl(packagesModal, packageName);
+
+      // check that bulk select and delete kebab is visible
+      await expect(packagesModal.locator('.pf-v6-c-table__td.pf-v6-c-table__check')).toBeVisible();
+      await expect(packagesModal.getByTestId('delete-kebab')).toBeVisible();
+
+      // click on the package to delete
+      const packageRow = packagesModal.getByRole('row').filter({ hasText: packageName });
+      await expect(packageRow).toBeVisible();
+      await row.getByRole('button', { name: 'Kebab toggle' }).click();
+      await page.getByRole('menuitem', { name: 'Delete package' }).click();
+
+      // expect that new delete packages? confirm modal shows up
+      await expect(page.getByTestId('delete_packages')).toBeVisible();
+      await expect(page.getByText('Delete packages?')).toBeVisible();
+      await expect(
+        page.getByTestId('confirm_delete_packages_table').getByText(packageName),
+      ).toBeVisible();
+
+      // confirm to delete the package
+      await Promise.all([
+        page.waitForResponse(
+          (resp) =>
+            resp.url().includes('/rpms/bulk_remove') && resp.status() >= 200 && resp.status() < 300,
+        ),
+        page.getByTestId('delete_packages_confirm').click(),
+      ]);
+
+      row = await getRowByNameOrUrl(page, uploadRepoName);
+      await waitForValidStatus(page, uploadRepoName);
+      await expect(row.getByTestId('package_count_button').getByText('0')).toBeVisible();
+
+      // verify that no packages show up in the package modal
+      packagesModal = await openPackagesModal(page, uploadRepoName);
+      await expect(packagesModal.getByRole('heading', { name: 'No packages' })).toBeVisible();
+      await packagesModal.getByRole('button').getByText('Close').click();
+    });
+  });
+
+  test('Non-upload repository does not show delete controls in PackageModal', async ({
+    page,
+    client,
+    cleanup,
+  }) => {
+    const externalRepoName = `${externalRepoNamePrefix}-${randomName()}`;
+
+    await cleanup.runAndAdd(() =>
+      cleanupRepositories(client, externalRepoNamePrefix, externalRepoUrl),
+    );
+    await closeGenericPopupsIfExist(page);
+    await navigateToRepositories(page);
+
+    await test.step('Create external snapshot repository', async () => {
+      // create external repository
+      await page.getByRole('button', { name: 'Add repositories' }).first().click();
+
+      await page.getByRole('textbox', { name: 'Name', exact: true }).fill(externalRepoName);
+      await page.getByLabel('Snapshotting').click();
+      await page.getByRole('textbox', { name: 'URL', exact: true }).fill(externalRepoUrl);
+      await page.getByRole('button', { name: 'Save', exact: true }).click();
+      await waitForValidStatus(page, externalRepoName);
+    });
+
+    await test.step('Verify delete controls are not available in PackageModal', async () => {
+      const packagesModal = await openPackagesModal(page, externalRepoName);
+
+      await expect(
+        packagesModal
+          .getByTestId('packages_table')
+          .getByRole('row')
+          .filter({ has: page.getByRole('gridcell') })
+          .first(),
+      ).toBeVisible();
+      await expect(
+        packagesModal
+          .getByTestId('packages_table')
+          .getByRole('row')
+          .filter({ has: page.getByRole('gridcell') })
+          .first()
+          .getByRole('button', { name: 'Kebab toggle' }),
+      ).toBeHidden();
+    });
+  });
+});
--- a/src/Pages/Repositories/ContentListTable/ContentListTable.tsx
+++ b/src/Pages/Repositories/ContentListTable/ContentListTable.tsx
@@ -502,12 +502,12 @@ const ContentListTable = () => {
   const pageSelectableRows = rows.filter((row) => !shouldDisableSelect(row));
 
   // How many of those selectable rows are currently selected
-  const pageSelectionCount = pageSelectableRows.filter(isSelected).length;
+  const numberOfRowsSelected = pageSelectableRows.filter(isSelected).length;
 
   const isPageSelected =
-    pageSelectableRows.length > 0 && pageSelectionCount === pageSelectableRows.length;
+    pageSelectableRows.length > 0 && numberOfRowsSelected === pageSelectableRows.length;
 
-  const isPagePartiallySelected = pageSelectionCount > 0 && !isPageSelected;
+  const isPagePartiallySelected = numberOfRowsSelected > 0 && !isPageSelected;
 
   const ouiaId = 'repositories-table';
 
@@ -539,6 +539,14 @@ const ContentListTable = () => {
     activeState === DataViewState.empty ||
     activeState === DataViewState.loading ||
     areNoSelectableRows;
+
+  const deleteRepoText = useMemo(
+    () =>
+      numberOfRowsSelected <= 1
+        ? 'Delete repository'
+        : `Delete ${numberOfRowsSelected} repositories`,
+    [numberOfRowsSelected],
+  );
 
   return (
     <>
@@ -698,9 +706,8 @@ const ContentListTable = () => {
                       contentOrigin.length === 0 ||
                       (isCommunityAndCustom && areNoSelectableRows)
                     }
-                    atLeastOneRepoChecked={pageSelectionCount > 0}
-                    numberOfReposChecked={pageSelectionCount}
-                    toggleOuiaId='repositories-kebab-toggle'
+                    atLeastOneRepoChecked={numberOfRowsSelected > 0}
+                    text={deleteRepoText}
                   />
                 </ConditionalTooltip>
               </FlexItem>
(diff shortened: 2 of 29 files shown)


xbhouse:
https://issues.redhat.com/browse/HMS-9614

src/Pages/Repositories/ContentListTable/components/SnapshotDetailsModal/Tabs/SnapshotPackagesTab.tsx
```diff
@@ -1,150 +1,73 @@
...
     isError,
     data = { data: [], meta: { count: 0, limit: 20, offset: 0 } },
-  } = useGetSnapshotPackagesQuery(snapshotUUID, page, perPage, debouncedSearchQuery);
+  } = useGetSnapshotPackagesQuery(snapshotUUID, page, perPage, debouncedSearch);
 
   useEffect(() => {
     if (isError) {
       onClose();
     }
   }, [isError]);
```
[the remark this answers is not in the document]
katarinazaprazna: Good catch by the bot here

src/services/Content/ContentApi.ts
```diff
@@ -188,6 +188,20 @@ export interface PackageItem {
   version: string;
 }
 
+export type PackageItemWithUUID = PackageItem & {
+  uuid: string;
+};
```
katarinazaprazna: I'm wondering, could we combine these into a single type `PackageItem` using a discriminated union based on the `uuid` field?

src/services/Content/ContentApi.ts
katarinazaprazna: Taking the approach ⬆️ means we can also clean up these extra response types. This should help make the types a bit easier to maintain. Let me know what you think

src/Pages/Repositories/ContentListTable/components/PackagesDeleteModal/PackagesDeleteModal.tsx
```diff
@@ -0,0 +1,178 @@
...
+  const { mutateAsync: removeRpms, isPending } = useBulkRemoveRepositoryRpmsMutate(
+    queryClient,
+    repoUUID,
+  );
+
+  useEffect(() => {
+    if (!selectedPackages.length) {
+      onClose();
+    }
+  }, [selectedPackages, onClose]);
```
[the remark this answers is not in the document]
katarinazaprazna: Good catch by the bot here

src/components/SharedTables/PackagesTable/usePackageTablePagination.ts
```diff
@@ -0,0 +1,18 @@
+import { useState } from 'react';
+
+export const usePackageTablePagination = ({ perPageKey }) => {
```
katarinazaprazna: Great call on creating a custom hook since the standard data-view one doesn't fit our paginated fetch!

I noticed `usePackageTablePagination` is super close to being generic. What do you think about renaming it (maybe something like `useTablePagination` or similar) so we can reuse it across other data-view tables down the line?

src/components/Tables/Packages/PackagesTableWithToolbars.tsx
```diff
@@ -1,88 +1,294 @@
...
+  });
 
-  const columnHeaders = ['Name', 'Version', 'Release', 'Architecture'];
+  // filter handlers
+  const { clearAllFilters, filters, onSetFilters } = filterProps;
+  const { setPage } = paginationProps;
+  const clearAllFiltersAndResetPage = useCallback(() => {
+    clearAllFilters();
+    setPage(1);
+  }, [clearAllFilters]);
```
katarinazaprazna, line 144: I believe `setPage` is missing from the dependency array

katarinazaprazna:
I noticed that in the empty state, there’s a bit of a gap between the left edge of the table and the start of the Name column. I wonder if the culprit might be the checkbox column?


Screenshot From 2026-05-19 12-53-52

src/components/DeleteKebab/DeleteKebab.tsx
```diff
@@ -7,12 +7,12 @@ import { EllipsisVIcon } from '@patternfly/react-icons';
 
 interface Props {
   atLeastOneRepoChecked: boolean;
-  numberOfReposChecked: number;
   toggleOuiaId?: string;
   isDisabled?: boolean;
+  text: string;
 }
 
-const DeleteKebab = ({ atLeastOneRepoChecked, numberOfReposChecked, isDisabled }: Props) => {
```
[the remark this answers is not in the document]
katarinazaprazna: Good catch by the bot here

_playwright-tests/UI/DeletePackages.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+  test('Non-upload repository does not show delete controls in PackageModal', async ({
+    page,
+    client,
+    cleanup,
+  }) => {
+    const externalRepoName = `${externalRepoNamePrefix}-${randomName()}`;
+
+    await cleanup.runAndAdd(() =>
+      cleanupRepositories(client, externalRepoNamePrefix, externalRepoUrl),
+    );
```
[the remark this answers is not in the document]
katarinazaprazna: Good catch by the bot here

src/components/SharedTables/PackagesTable/index.tsx
```diff
@@ -1,88 +1,294 @@
...
+  enabledRowActions?: boolean;
+  paginationProps: PaginationProps;
+  filterProps: FilterProps;
+  selection?: ReturnType<typeof useDataViewSelection>;
 }
 
-export default function PackagesTable({
-  isFetchingOrLoading,
-  isLoadingOrZeroCount,
+export const usePackagesTableSelection = () => {
```
katarinazaprazna: I wonder if it's worth keeping the `usePackagesTableSelection` abstraction here. We might be able to just use `useDataViewSelection` directly to keep things a bit leaner. What do you think?

src/Pages/Repositories/ContentListTable/components/PackageModal/PackageModal.tsx
```diff
@@ -1,184 +1,141 @@
...
+export const useEnableDelete = (repoUUID) => {
+  const { data: repository, isError } = useFetchContent(repoUUID);
+
+  const isUploadRepository = repository?.origin === ContentOrigin.UPLOAD;
+
+  return [isUploadRepository, isError];
+};
+
+export default function PackageModal() {
+  const { repoUUID } = useParams();
```
katarinazaprazna: I realised we actually have a `useSafeUUIDParam` hook that we don't use too often! What do you think about making use of it? It would save us from having to do that extra type cast (L63)

katarinazaprazna:
This definitely isn't a blocker, but for unification purposes, would you be open to renaming `PackagesTableWithToolbars` to `PackagesTable`? Along with that, we could rename `PackagesBasicTable` to `DeletePackagesTable` and move it to its own file. That feels pretty aligned with how we handle tables elsewhere in the repo, but totally up to you!

src/Pages/Templates/TemplateDetails/components/Tabs/TemplatePackageTab.tsx
```diff
@@ -1,135 +1,52 @@
...
-  const fetchingOrLoading = isFetching || isLoading;
-
-  const loadingOrZeroCount = fetchingOrLoading || !count;
+  const paginationProps = {
+    itemCount: count,
+    ...paginationData,
+  };
 
   if (isLoading) {
     return <Loader />;
```
katarinazaprazna: I wonder if the `Loader` component is needed here? I believe the underlying data-view table is wired to handle the loading state when `isLoading` or `isFetching` are provided (`DataViewState.loading`)

src/services/Content/ContentApi.ts
Starle21: Keeping both types for now, since the narrower type is used in PackagesDeleteModal and received from backend with getPackages

src/services/Content/ContentApi.ts
Starle21: Ditto the comment :arrow_up:

src/components/SharedTables/PackagesTable/usePackageTablePagination.ts
Starle21: Agreed! I created a new folder under components called Tables, where all the various tables can be gathered. 
Then inside there are folders such as Generic for the most standard stuff and Package for everything specific for PackageTables. Similar pattern can be followed with other types of tables

src/Pages/Repositories/ContentListTable/components/PackageModal/PackageModal.tsx
Starle21: Agreed, applied in the changed files.

src/Pages/Templates/TemplateDetails/components/Tabs/TemplatePackageTab.tsx
Starle21: Great catch!

src/components/SharedTables/PackagesTable/index.tsx
Starle21: The idea is to keep any DataView stuff out of the consuming components, so they don't see any DataView. I will keep this for now.

Starle21:
To generalize tables a bit, as we have discussed, I created folder Tables in components.
Under Generic, there is stuff that can be reused over different types of tables.
Under Packages, there is stuff specifically for Packages tables only.

I tried to make something similar to `useAdvisoriesTableState`, but I found it hard to do due to the need to use different individual parts such as Pagination, Filters and such in consuming components.

src/components/Tables/Packages/PackagesTableWithToolbars.tsx
Starle21: React states that setter functions from useState have a stable identity across re-renders, therefore setPage does not have to be in the dependency array.

src/Pages/Repositories/ContentListTable/components/PackagesDeleteModal/PackagesDeleteModal.tsx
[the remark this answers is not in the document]
Starle21: There is always at least one selected package. This is only a safe precaution If somehow that modal should be rendered directly and nothing is selected - therefore it should redirect away, as no confirm can be made.

Starle21:
> I noticed that in the empty state, there’s a bit of a gap between the left edge of the table and the start of the Name column. I wonder if the culprit might be the checkbox column?
> 
Fixed.
The issue was that the containing DataView component did not catch the fact that in the empty state there is no checkbox for a selection and was keeping it in there.

src/components/Tables/Generic/hooks/useTablePaginationLocalStorage.ts
```diff
@@ -0,0 +1,35 @@
...
+  ) => void;
+  setPage: (page: number) => void;
+};
+
+export const useTablePaginationLocalStorage = ({ key }: TablePaginationLocalStorage) => {
+  const [page, setPage] = useState(1);
+  const storedPerPage = Number(localStorage.getItem(key)) || 20;
+  const [perPage, setPerPage] = useState(storedPerPage);
+
+  const onPerPageSelect = (_, newPerPage, newPage) => {
```
katarinazaprazna: Thanks for putting the pagination logic into a generic hook! 

Just a small suggestion: we might want to wrap the functions in `useCallback` so they don't recreate on every render. We could even expose `setPage` directly

src/components/Tables/Generic/hooks/useTablePaginationLocalStorage.ts
```diff
@@ -0,0 +1,35 @@
...
+    e: React.MouseEvent | React.KeyboardEvent | MouseEvent,
+    perPage: number,
+    newPage: number,
+  ) => void;
+  setPage: (page: number) => void;
+};
+
+export const useTablePaginationLocalStorage = ({ key }: TablePaginationLocalStorage) => {
+  const [page, setPage] = useState(1);
+  const storedPerPage = Number(localStorage.getItem(key)) || 20;
```
katarinazaprazna: Just a tiny note on performance: passing the function directly like `useState(() => ...)` ensures it only executes on the first render rather than running the `localStorage` check on every re-render

src/components/Tables/Generic/hooks/useTableSelectionOnID.ts
```diff
@@ -0,0 +1,6 @@
+import { useDataViewSelection } from '@patternfly/react-data-view/dist/dynamic/Hooks';
+
+export const useTableSelectionOnID = () => {
```
katarinazaprazna: To keep things simple, we could skip this wrapper and call `useDataViewSelection` directly. What do you think about bypassing it?

src/Pages/Repositories/ContentListTable/components/PackagesDeleteModal/PackagesDeleteModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const onClose = () =>
+    navigate(
+      `${rootPath}/${REPOSITORIES_ROUTE}` +
+        (contentOrigin.length === 1 && contentOrigin[0] === ContentOrigin.REDHAT
+          ? `?origin=${contentOrigin}`
+          : ''),
+    );
+
+  return onClose;
+};
```
katarinazaprazna: As both of these hooks are only used once and defined in the same file, we could probably simplify things by skipping them entirely and inlining the logic directly into `PackagesDeleteModal`. This can make the code easier to follow

src/Pages/Repositories/ContentListTable/components/PackagesDeleteModal/PackagesDeleteModal.tsx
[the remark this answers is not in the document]
katarinazaprazna: I see, thanks for the explanation!

src/Pages/Repositories/ContentListTable/ContentListTable.tsx
```diff
@@ -540,6 +540,15 @@ const ContentListTable = () => {
...
     areNoSelectableRows;
 
+  const deleteRepoText = useMemo(() => {
+    const numberOfReposChecked = pageSelectionCount;
+    const text =
+      numberOfReposChecked <= 1
+        ? 'Delete repository'
+        : `Delete ${numberOfReposChecked} repositories`;
+    return text;
+  }, [pageSelectionCount]);
```
katarinazaprazna: What do you think about putting this ternary directly in the JSX? Since it’s a quick check, we could skip the intermediate variable and memoization altogether

src/services/Content/ContentApi.ts
```diff
@@ -188,6 +188,15 @@ export interface PackageItem {
   version: string;
 }
```
katarinazaprazna: Just a tiny tweak to keep the public API clean: What do you think about removing the export here? It's an intermediate type only used inside `ContentApi.ts`

katarinazaprazna:
Great work overall, and thanks for jumping on the previous feedback!

I've left a few comments to look over. The ones on `useTablePaginationLocalStorage` and `useTableSelectionOnID` are key since they affect all our tables. The rest are suggestions for simplicity and performance that you can apply across the remaining files where relevant. I only commented on a few spots, but you might spot the same patterns in a couple of other files too. Please let me know if anything is unclear, happy to discuss!

src/components/Tables/Generic/hooks/useTablePaginationLocalStorage.ts
Starle21: Sure thing, added useCallback.
Both `setPage` and `setPerPage` have been exposed directly, since they are memoized by React.

src/components/Tables/Generic/hooks/useTablePaginationLocalStorage.ts
Starle21: Yes, got rid of that line.

src/Pages/Repositories/ContentListTable/components/PackagesDeleteModal/PackagesDeleteModal.tsx
Starle21: I prefer to keep both hooks extracted out. 

It depends on what you want to optimize the code "to be easier to read" for.
Here, the routing details are really details, the "meat" of the component is the table of packages to delete and the final confirmation which sends a signal to the backend to delete them.

Routing out of the modal is really a secondary matter. In my view, if there is less code with details grouped and encapsulated into its own function, it makes the code easier to read. The details about routing are optional, you only dig into them when you care.

You can also think about this as the "single responsibility principle", the function only cares about navigation; the parent component only about composing multiple things to provide delete package confirmation modal. It is actually similar thinking as in extracting `useTablePagination`.

The exact same code for navigating to repositories is in `PackageModal`, so extracting that into own hook file under `Hooks/navigation`. Also I've seen in couple of other places similar code, so it might pay off to group it together under that navigation folder.

src/services/Content/ContentApi.ts
Starle21: Done.

src/Pages/Repositories/ContentListTable/ContentListTable.tsx
Starle21: Well, a better practice in general is to "prepare data" outside jsx, especially if it is anything beyond a trivial boolean or one-liner.

JSX is a declarative structure, so it it is better to name the intent (`deleteRepoText`) outside and then use that in the JSX, so it reads as a "what" is the structure and the "how" is above.

This way it is easier also to divide the big component later between "presentational components", which just accept data in parameters, and "container components" which know how to talk with the backend and "prepare" data for the presentational components downstream.

Also if there si any change in the logic - it is easier to do it in this named value, inline JSX does not scale.

src/components/Tables/Generic/hooks/useTableSelectionOnID.ts
Starle21: I would keep it this way, as DataView is still quite new, and it is easier to make changes in one file than in many, logically the consuming component should not care about any DataView stuff and matching logic, but I do not have a strong opinion, so I inlined it.

src/Pages/Repositories/ContentListTable/components/PackagesDeleteModal/PackagesDeleteModal.tsx
katarinazaprazna: Thanks for flagging that this is a repeated pattern, I wasn't aware of that! I looked into it and found the same navigation logic duplicated across at least 6 components, so extracting it is definitely the right call

One thought. Since `useNavigateToRepositories` and `useNavigateToRoot` only differ by whether `/${REPOSITORIES_ROUTE}` is appended to the path, we could consolidate them into a single hook with an optional route parameter. That way there's one hook, one place to maintain the origin logic, and callers just pass `REPOSITORIES_ROUTE` or nothing. What do you think?

src/Pages/Repositories/ContentListTable/ContentListTable.tsx
katarinazaprazna: I absolutely agree with the general principle of extracting complex data preparation out of JSX. That said, this specific case is a one-liner ternary:

```
pageSelectionCount <= 1 ? 'Delete repository' : `Delete ${pageSelectionCount} repositories`
```

In my opinion, wrapping it in `useMemo` with a dependency array adds more code than the logic itself. I don't have a strong preference on this though, happy to leave it

src/Pages/Repositories/ContentListTable/components/PackagesDeleteModal/PackagesDeleteModal.tsx
Starle21: Ok, created a record which keeps `destinationKey: function that builds location string type' with the hook useNavigateTo. Components just pass destinationKey and are good to go.

katarinazaprazna (approved):
Thanks you so much for the changes!

PR #1010, opened 2026-05-15
HMS-10509: dark mode compatibility fixes

dominikvagner:
## Summary
This makes two changes for dark mode compatibility:
- It applies the default background for floating elements to tables and paginations inside of modals, so that they have the same look as in light mode. _(this solution will also work for data view tables down the line)_
- This changes the token used for status colors to the default version which change with the applied theme. Also makes the text uncolored per UX request.

> ⬆️ IMO, we might wanna revisit this later, at the upcoming SPUR maybe, but these are the proposed changes by UX currently 😅 

_Additionally, makes a one-liner fix for the snapshot selector so it's scrollable and can be actually used for repos with a lot of snapshots._

## Testing steps
- Verify that all tables, that are inside of modals (raw modals, not wizard) look good 💅🏼 
- Check that status colors are now readable/visible in dark mode 👓

### DIFF
--- a/src/Pages/Repositories/AdminTaskTable/components/StatusIcon.tsx
+++ b/src/Pages/Repositories/AdminTaskTable/components/StatusIcon.tsx
@@ -7,16 +7,16 @@ import {
 import { Flex, FlexItem, Spinner } from '@patternfly/react-core';
 import StatusText from 'components/StatusText/StatusText';
 import {
-  t_global_color_status_danger_100 as global_danger_color_100,
-  t_global_color_status_success_100 as global_success_color_100,
-  t_global_color_status_warning_100 as global_warning_color_100,
+  t_global_color_status_danger_default,
+  t_global_color_status_success_default,
+  t_global_color_status_warning_default,
 } from '@patternfly/react-tokens';
 import { createUseStyles } from 'react-jss';
 import { AdminTask } from 'services/Admin/AdminTaskApi';
 
-const red = global_danger_color_100.value;
-const green = global_success_color_100.value;
-const gold = global_warning_color_100.value;
+const red = t_global_color_status_danger_default.var;
+const green = t_global_color_status_success_default.var;
+const gold = t_global_color_status_warning_default.var;
 
 const useStyles = createUseStyles({
   spinner: {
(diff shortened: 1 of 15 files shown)


xbhouse:
https://issues.redhat.com/browse/HMS-10509

src/Pages/Repositories/AdminTaskTable/components/StatusIcon.tsx
xbhouse, line 66: i think the imported colors for these icons will need to be switched over to the default ones too. only "admins" see this but just to be consistent :)

src/Pages/Repositories/ContentListTable/components/SnapshotListModal/components/ChangedArrows.tsx
```diff
@@ -1,8 +1,8 @@
 import { LongArrowAltDownIcon, LongArrowAltUpIcon } from '@patternfly/react-icons';
 import { Flex, FlexItem } from '@patternfly/react-core';
 import {
-  t_global_color_status_danger_100,
-  t_global_color_status_success_100,
+  t_global_color_status_danger_default,
```
xbhouse, line 4: do all of the `danger` tokens look the same in dark and light mode or is my eyesight just terrible? in the [patternfly docs](https://www.patternfly.org/components/button) in dark mode, it looks to me more orange but the same tokens are used in our app 🤔

xbhouse (commented):
thank you for this (and for fixing the snapshot selector)! this looks good, i have just a couple comments 😄 

aside from those, from what i can tell it looks like the right tokens are used, so if there are any contrast issues it might be platform-wide. i agree we should bring this up and revisit the status text color with UX 👍

src/Pages/Repositories/ContentListTable/components/SnapshotListModal/components/ChangedArrows.tsx
dominikvagner: well I see them differently in light and dark modes 👀😆
it looking more orange is just a contrast visual "trick" I'd guess (the color swatch in the docs is much bigger so it looks brighter and also it not being and icon means it's a fully solid color) 🪄 
if I take a look at them with a color picker extension I get the same color codes 😄🔍

src/Pages/Repositories/AdminTaskTable/components/StatusIcon.tsx
dominikvagner: oops, missed that one, fixed 🫡 😄

rverdile (approved):

PR #1023, opened 2026-05-28
Build: update `yarn.lock` for version package alignment (data-view)

katarinazaprazna:
## Summary

PatternFly has released data-view v6.5.0 as a stable version. We have switched from the prerelease version to this stable release, as we no longer require any features specific to the prerelease.

### Changes

- Manually triggered the update of `yarn.lock` to ensure proper package version alignment, as it seems that Dependabot missed this during the initial switch


swadeley (approved):
ACK

PR #1030, opened 2026-06-05
Test: Update template tests cleanup sequence

swadeley:
## Summary
    Update template tests cleanup sequence
    We want clients cleaned before template

## Testing steps
tests pass

### DIFF
--- a/_playwright-tests/Integration/AssignEUSTemplateToSystem.spec.ts
+++ b/_playwright-tests/Integration/AssignEUSTemplateToSystem.spec.ts
@@ -44,13 +44,13 @@ test.describe('Assign EUS Template to System', () => {
     void client; // Pull in fixture so Undici fetch dispatcher is configured for dynamic API cleanup
 
     await test.step('Set up cleanup for templates and RHSM client', async () => {
+      cleanup.add(() => regClient.Destroy('rhc'));
       await cleanup.runAndAdd(async () => {
         await ensureValidToken(page, 'EUS_REPO_TOKEN.json', 5);
         const apiBasePath = process.env.BASE_URL + '/api/content-sources/v1';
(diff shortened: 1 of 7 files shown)


dominikvagner (commented):
this seems like a wrong solution :/ 
it adds new slightly different way of running cleanups, which complicates things, and the cleanups still in the end are run in parallel I think 🤔 

if the parallel cleanup complicates things, let's make them sequential in the cleanup fixture 💭 image builder has that [here](https://github.com/osbuild/image-builder-frontend/blob/80c4f2b8a4f840d416d3b96994f2685a44ba7403/playwright/fixtures/cleanup.ts#L36-L45) and change to that in our test utils [here](https://github.com/content-services/content-sources-backend/blob/da5f1b949927dbd19aaac7a1c32b30e9eb248860/_playwright-tests/test-utils/src/fixtures/cleanup.ts#L63-L68)

swadeley:
> this seems like a wrong solution :/ it adds new slightly different way of running cleanups, which complicates things, and the cleanups still in the end are run in parallel I think 🤔
> 
> if the parallel cleanup complicates things, let's make them sequential in the cleanup fixture 💭 image builder has that [here](https://github.com/osbuild/image-builder-frontend/blob/80c4f2b8a4f840d416d3b96994f2685a44ba7403/playwright/fixtures/cleanup.ts#L36-L45) and change to that in our test utils [here](https://github.com/content-services/content-sources-backend/blob/da5f1b949927dbd19aaac7a1c32b30e9eb248860/_playwright-tests/test-utils/src/fixtures/cleanup.ts#L63-L68)

Hi @dominikvagner 

I have done what you suggested here:  https://github.com/content-services/content-sources-backend/pull/1532

I think we should cleanup clients before the template they are attached to so I am using this PR to change the sequence we call the cleanup.

dominikvagner (approved):
ack ✅

PR #1031, opened 2026-06-05
Build: extend malware detection for PRs

katarinazaprazna:
## Summary

Our Aikido malware feed was previously integrated into GitHub's dependency-review-action. However, a [bug causes that action to ignore namespaces and strip scopes on npm packages](https://github.com/actions/dependency-review-action/issues/1014), triggering false positives that forced us to temporarily disable the malware blocklist in our PR checks.

To fix this:

- Refactored the original malware script to extract newly added dependencies by diffing `yarn.lock`. It matches both namespaces and transitive dependencies against the Aikido feed
- Kept the `actions/dependency-review-action` to scan for traditional vulnerabilities

The security gates run as separate jobs so both report independently on the PR. The idea is for PR authors see all security issues in one pass instead of fixing one, re-pushing, and only then discovering the other.


.github/workflows/content-sources-actions.yml
```diff
@@ -133,16 +133,16 @@ jobs:
...
-          chmod +x workflowScripts/malware-to-purl.py
-          python3 workflowScripts/malware-to-purl.py
+          chmod +x workflowScripts/malware-pr-check.py
+          python3 workflowScripts/malware-pr-check.py
 
-      - name: Dependency Review
+      - name: Known vulnerability review
         uses: actions/dependency-review-action@v4
-        with:
-          config-file: .github/dependency-review-config.yml
```
dominikvagner, line 145: looking at the dependency review action there seems to be a v5 👀  
let's upgrade that so we don't forget 😅 

and the old script also created the config, since the denied pacakges option is a bit broken, we can skip that, but we probably should include some of the options 💭 
these I would keep so it's not too spammy, but alerts us on stuff we want: `comment-summary-in-pr: on-failure, vulnerability-check: true, license-check: false` _(these can be inlined in the workflow [[1](https://github.com/actions/dependency-review-action#option-1-using-inline-configuration)])_

.github/workflowScripts/malware-pr-check.py
```diff
@@ -0,0 +1,110 @@
...
+        sys.exit(0)
+
+    added_packages = parse_added_packages(diff_text)
+    print(f"Found {len(added_packages)} unique packages added or updated in this PR.")
+
+    # Compare and enforce
+    compromises = []
+    for pkg in added_packages:
+        if pkg in malware_packages:
+            compromises.append(pkg)
```
dominikvagner: does this handle version at all? 🤔 if not then this could produce a LOT of false positives 🙅🏼 

and the script doesn't seem to check just version bumps 👀  (which makes sense if it doesn't check version, but then it might miss a lot), this ⬇️  would be ignored, which could be the bad version 💭 
```diff
 tmp@^0.2.0, tmp@^0.2.5:
-  version "0.2.5"
-  resolved "https://registry.yarnpkg.com/tmp/-/tmp-0.2.5.tgz#b06bcd23f0f3c8357b426891726d16015abfd8f8"
-  integrity sha512-voyz6MApa1rQGUxT3E+BK7/ROe8itEx7vD8/HEvt4xwXucvQ5G5oeEiHkmHZJuBO21RpOf+YYm9MOivj709jow==
+  version "0.2.7"

```

.github/workflows/content-sources-actions.yml
katarinazaprazna: Good catch, thanks! I'll bump the version. I was also a bit worried about the excessive alerting here. Will tweak and push an update shortly

.github/workflowScripts/malware-pr-check.py
katarinazaprazna: Great point. Will update this too!

.github/workflowScripts/malware-pr-check.py
```diff
@@ -0,0 +1,132 @@
...
+
+    return added_packages
+
+def report_compromises(malware_packages, added_packages):
+    """Finds overlap between added and malware packages."""
+    base, check = (
+        (added_packages, malware_packages)
+        if len(added_packages) >= len(malware_packages)
+        else (malware_packages, added_packages)
+    )
```
dominikvagner: this breaks the logic in the compromises version check below 🥲 
let's just do `(added_packages, malware_packages)`, I don't think the performance should matter that much here, but if you wanted to make it speedy 🚗 
the malware packages could be turned into a map where the package name would be a key and value would be the version(s), lookups should be faster 💭

dominikvagner (changes requested):
small bug in the compromises check 🐛 

and what we never did correctly was blocking all other actions (or at least the ones which install npm stuff) if this fails 💭 
we should probably do that tho, could you take a look please 😅

.github/workflowScripts/malware-pr-check.py
katarinazaprazna: Thanks for catching these! I'll dive in and get both issues sorted out shortly 🏗️

.github/workflowScripts/malware-pr-check.py
```diff
@@ -0,0 +1,138 @@
...
+            for name in current_names:
+                added_packages.add((name, version))
+            current_names = []
+
+    return added_packages
+
+
+def report_compromises(malware_packages, added_packages):
+    """Finds overlap between added and malware packages."""
+    compromises = added_packages.intersection(malware_packages)
```
dominikvagner: this won't detect malware packages that are reported with all versions as malicious 🐛 i.e.: `('pkg_name', '*')` 
just a basic loop is the way here 😅

dominikvagner (changes requested):

dominikvagner (approved):
looks good ✅ 

_sidenote: 💭 one thing we could also look at is making the playwright workflows dependent on the malware check, but we can do that in a different PR 😅_

PR #1033, opened 2026-06-09
Build: Add agent rules for github actions

swadeley:
## Summary

Add agent rules for github actions

Working with forks presents extra challenges in using keys and tokens in Github actions.
This doc should help AI agents assist users write Github actions.

## Testing steps
tests pass.

### DIFF
--- a/docs/rules/github-actions.md
+++ b/docs/rules/github-actions.md
@@ -1,0 +1,32 @@
+# GitHub Actions, secrets, and forked pull requests
+
+## Secrets made available as environment variables
+
+Secrets used in GitHub Actions workflows are stored in a shared secret store and not visible to agents. They are made available in CI as environment variables. Contact someone in the content-sources team for guidance.
+
+
+## Secrets and variables on fork PRs
+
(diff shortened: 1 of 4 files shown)


docs/rules/github-actions.md
```diff
@@ -0,0 +1,37 @@
+# GitHub Actions, secrets, and forked pull requests
+
+## AWS environment variables
```
xbhouse: the fork PR section is really helpful 🙂 since the `AWS environment variables` section doesn't give agents anything really actionable, would it make sense to drop that section and keep this focused on CI behavior in forks? 

alternatively, if we do want an intro or to prevent agents from searching for these values, maybe something like this would be enough: `Secrets used in GitHub Actions workflows are stored in a shared secret store and not visible to agents. Contact someone in the content-sources org/team for guidance.`

swadeley: Hi @xbhouse , thank you for you review. Changes made.

xbhouse (approved):

PR #1049, opened 2026-06-22
HMS-10844: refactor useNavigateTo where appropriate

Starle21:
## Summary
- makes use of useNavigateTo hook for frequently used navigation paths - such as `repositories` or `repositorySnapshots`
- gathering the paths in one place for easy change
- cleaning up code a bit
- adding unit tests for useNavigateTo hook to make sure it returns the right paths

## Testing steps
- testing manually: all the navigation from the components where the changes happened should work as it previously did
- all tests pass

### DIFF
--- a/src/Pages/Repositories/ContentListTable/components/SnapshotDetailsModal/SnapshotDetailsModal.test.tsx
+++ b/src/Pages/Repositories/ContentListTable/components/SnapshotDetailsModal/SnapshotDetailsModal.test.tsx
@@ -1,10 +1,9 @@
-import { render, screen, waitFor, within } from '@testing-library/react';
+import { render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import SnapshotDetailsModal, { SnapshotDetailTab } from './SnapshotDetailsModal';
-import { useAppContext } from 'middleware/AppContext';
-import { ContentOrigin } from 'services/Content/ContentApi';
+import { useNavigateTo } from 'Hooks/navigation/useNavigateTo';
 
-const mockNavigate = jest.fn();
+const mockUseNavigateTo = useNavigateTo as jest.Mock;
 const mockSetSearchParams = jest.fn();
 
 jest.mock('./Tabs/SnapshotPackagesTab', () => ({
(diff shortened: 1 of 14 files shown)


xbhouse:
https://issues.redhat.com/browse/HMS-10844

src/Pages/Repositories/ContentListTable/components/SnapshotDetailsModal/SnapshotDetailsModal.test.tsx
```diff
@@ -1,10 +1,14 @@
...
-import { ContentOrigin } from 'services/Content/ContentApi';
 
-const mockNavigate = jest.fn();
+const mockNavigateToRepositories = jest.fn();
+const mockNavigateToSnapshots = jest.fn();
+const mockUseNavigateTo = jest.fn((key: string) => {
+  if (key === 'repositories') return mockNavigateToRepositories;
+  if (key === 'repositorySnapshots') return mockNavigateToSnapshots;
+  return jest.fn();
+});
```
katarinazaprazna: Awesome work setting up a separate unit test for the `useNavigateTo` hook! Since that's in place, we can simplify this mock to just check if the component calls it with the right keys. No need to re-implement the routing logic, just verifying the integration works

```
jest.mock('Hooks/navigation/useNavigateTo', () => ({
  useNavigateTo: jest.fn(),
}));
```

katarinazaprazna (dismissed):
Great work on this PR! The code is much cleaner now. I've left one non-blocking comment about simplifying one of the tests since we have coverage elsewhere

Awesome job overall 🚀

src/Pages/Repositories/ContentListTable/components/SnapshotDetailsModal/SnapshotDetailsModal.test.tsx
Starle21: Simplified the test to its bare, testing that it only is called with the right keys. Good catch! :+1:

katarinazaprazna: Great work on this, thanks for tackling it! The test logic is much cleaner now

katarinazaprazna (approved):

PR #1063, opened 2026-07-03
Lightwell: Hide submenu items that can get you away from Lightwell

marusak:
We want to make sure that customer cannot jump away from Lightwell if there is no way how to navigate back. Also we want to keep it as simple as possible not to distract customers with options that are not relevant for Lightwell.


TenSt (approved):
lgtm!

PR #1077, opened 2026-07-06
Lightwell: repositories page UI improvements

arburka:
## Summary
- Add info tooltips to Packages and Builds column headers explaining the distinction
- Bold repo name links for visual hierarchy
- Reduce gap between page header and table
- Balance table card top/bottom padding

## Test plan
- [ ] Navigate to `/lightwell`
- [ ] Hover over Packages/Builds column headers to see info tooltips (no duplicate tooltip on header text)
- [ ] Repo names should be bold
- [ ] Spacing between header and table should be tight
- [ ] Table card should have visually balanced top/bottom padding

co-authored by Claude Code

### DIFF
--- a/src/Pages/Lightwell/Repositories/RepositoriesTable.tsx
+++ b/src/Pages/Lightwell/Repositories/RepositoriesTable.tsx
@@ -97,7 +97,11 @@ const RepositoriesTable = () => {
 
   if (isError) throw error;
 
-  const columnHeaders: { title: string; width?: BaseCellProps['width']; info?: ComponentProps<typeof Th>['info'] }[] = [
+  const columnHeaders: {
+    title: string;
+    width?: BaseCellProps['width'];
+    info?: ComponentProps<typeof Th>['info'];
+  }[] = [
     { title: 'Repository' },
     { title: 'Ecosystem', width: 15 },
     { title: 'Security level', width: 15 },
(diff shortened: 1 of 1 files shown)


src/Pages/Lightwell/Repositories/RepositoriesTable.tsx
```diff
@@ -96,12 +98,23 @@ const RepositoriesTable = () => {
 
   if (isError) throw error;
 
-  const columnHeaders: { title: string; width?: BaseCellProps['width'] }[] = [
+  const columnHeaders: { title: string; width?: BaseCellProps['width']; info?: ThInfoType }[] = [
```
xbhouse: since we're already using `Th`, what do you think about retrieving the info type from the component instead of importing `ThInfoType`?

something like this:
```
const columnHeaders: {
  title: string;
  width?: BaseCellProps['width'];
  info?: ComponentProps<typeof Th>['info'];
}[] = [
```
this would avoid depending on `ThInfoType` being exported, which doesn't seem to be the case in our current PatternFly version

xbhouse (commented):
this looks great otherwise :) thank you for the PR!

arburka:
Thanks for checking it out @xbhouse!

xbhouse (commented):
could you also run `yarn lint:fix` to fix the lint error?

xbhouse (approved):
thank you!

PR #1078, opened 2026-07-06
Lightwell: packages list page UI improvements

arburka:
## Summary
- Bold package name links for visual hierarchy (consistent with repos page)
- Remove bottom padding from description to tighten header spacing
- Reduce titleWrapper top padding from 24px to 16px for even breadcrumb/title/description spacing
- Use default (non-compact) size for repo URL label
- Add copy feedback tooltip to repo URL label ("Click to copy URL" / "Copied!")
- Fix search placeholder to say "Filter by name" (namespace search not supported by API)
- Use relative timestamps ("4 days ago") via dayjs + PF Timestamp with UTC tooltip on hover

## Test plan
- [ ] Navigate to `/lightwell/java-remediated`
- [ ] Package names should be bold
- [ ] Header spacing should be even between breadcrumb, title, and description
- [ ] Repo URL label should show tooltip on hover, "Copied!" on click
- [ ] Search placeholder should say "Filter by name"
- [ ] Last updated column should show relative time with full date tooltip on hover
- [ ] Font size in Last updated column should match rest of table

co-authored by Claude Code

PR #1081, opened 2026-07-07
Lightwell: Use copy button consistently

marusak:
Copy buttons do not provide any feedback when clicked. Also Labels were used and not the ClipboardCopy components.

This changes the visual slightly - the copy icon is on the right and not left and only the icon is clickable and not the whole label. I had to fake round edges with css - not thrilled about it but I feel that the feedback when clicked was worth it.

@xbhouse Please feel free to close if that was done on purpose.

Before:
Screenshot From 2026-07-07 10-36-57

After:

Screenshot From 2026-07-07 10-31-52

### DIFF
--- a/src/Pages/Lightwell/Packages/PackagesTable.tsx
+++ b/src/Pages/Lightwell/Packages/PackagesTable.tsx
@@ -3,12 +3,13 @@ import {
   BreadcrumbItem,
   Button,
   Card,
+  ClipboardCopy,
+  ClipboardCopyVariant,
   Content,
   Flex,
   FlexItem,
   Grid,
   Icon,
-  Label,
   Pagination,
   PaginationVariant,
   SearchInput,
@@ -19,7 +20,7 @@ import {
   ToolbarContent,
   ToolbarItem,
 } from '@patternfly/react-core';
-import { CopyIcon, CodeIcon, JavaIcon, PythonIcon } from '@patternfly/react-icons';
+import { CodeIcon, JavaIcon, PythonIcon } from '@patternfly/react-icons';
 import { SkeletonTable } from '@patternfly/react-component-groups';
 import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
 import { createUseStyles } from 'react-jss';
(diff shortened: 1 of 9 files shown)


xbhouse (dismissed):

styles/lightwell-clipboard-copy.scss
```diff
@@ -0,0 +1,10 @@
+.pf-v6-c-clipboard-copy__expandable-content,
+.pf-v6-c-clipboard-copy__group,
+.pf-v6-c-clipboard-copy.pf-m-inline {
+  border-radius: var(--pf-t--global--border--radius--pill);
```
xbhouse, line 2: looks like this changes the border radius of the popover code snippet too and cuts off some of the text inside

xbhouse: ah the maven coordinate release on the package details header is a little truncated too

xbhouse: maybe the changes in this [PR](https://github.com/content-services/content-sources-frontend/pull/1078) on the repo URL match what you're after?

marusak: > looks like this changes the border radius of the popover code snippet too

Doh, yes! Good catch. Fixed.

> ah the maven coordinate release on the package details header is a little truncated too

Not sure what you mean :thinking: 

> maybe the changes in this https://github.com/content-services/content-sources-frontend/pull/1078 on the repo URL match what you're after?

I think it only adds the click tooltip to self-standing copy buttons but not to the inline grey ones

TenSt (dismissed):
lgtm!

styles/lightwell-clipboard-copy.scss
xbhouse: i mean here:
Screenshot From 2026-07-07 09-39-08

marusak: I just properly read through @arburka changes and he indeed adds these popups manually. Feels a bit hacky, but :man_shrugging: Let me comment in his PR.

And I can review the change you mentioned.

src/Pages/Lightwell/Packages/PackagesTable.tsx
```diff
@@ -132,14 +134,16 @@ const PackageCopyLabel = ({ name, groupId, version, isPython }: PackageCopyLabel
...
+      isReadOnly
+      hoverTip='Copy'
+      clickTip='Copied'
+      variant={ClipboardCopyVariant.inlineCompact}
+      copyAriaLabel={`Copy ${version}`}
+      onCopy={() => navigator.clipboard.writeText(copyText)}
     >
       {version}
-    </Label>
+    </ClipboardCopy>
```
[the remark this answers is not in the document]
arburka: Visually looks good - just happy to have the feedback when clicked. Thanks!

styles/lightwell-clipboard-copy.scss
marusak: Ok, so that one instance you mentioned was not Label but Button. Styling CopyClipboard to look like secondary button would be too much of a workaround, so added it back and gave it tooltip it deserves.

rverdile (approved):

PR #1087, opened 2026-07-07
Test: Add debug playwright workflow

swadeley:
## Summary
Add debug playwright workflow

Enables running the same test multiple times with given interval.
Captures console logs and saves them in machine readable format.
Intended for manual and AI assisted debugging of flaky tests.
## Testing steps

#testwith https://github.com/content-services/content-sources-backend/pull/1580

### DIFF
--- a/.github/workflowScripts/runPlaywrightDebug.sh
+++ b/.github/workflowScripts/runPlaywrightDebug.sh
@@ -1,0 +1,179 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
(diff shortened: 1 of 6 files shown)


dominikvagner (approved):
the script seems to be very defensively written, maybe a bit too much for a debugging script _(LLMs do that sometimes)_ 😆 but for a debugging side utility it's okay imo 😄 

thanks for the test fixes! 🙇🏼 

ack! ✅

PR #1098, opened 2026-07-10
HMS-9840: add RepositoriesCard component

Starle21:

### DIFF
--- a/_playwright-tests/Integration/NoSubsUserTest.spec.ts
+++ b/_playwright-tests/Integration/NoSubsUserTest.spec.ts
@@ -5,7 +5,11 @@ import {
   USING_CONTENT_TEMPLATES_PAGE,
 } from 'constants/docs';
 import { closeGenericPopupsIfExist, waitForValidStatus } from '../UI/helpers/helpers';
-import { getZeroStateTitleLocator } from '../UI/helpers/navHelpers';
+import {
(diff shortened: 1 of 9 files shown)


xbhouse:
https://issues.redhat.com/browse/HMS-9840

swadeley:
Hi @Starle21 , 
from Templates page I select  "Red Hat repositories", I arrive at the repo page with the correct filters selected.
I see URL is now `https://stage.foo.redhat.com:1337/insights/content/repositories?origin=red_hat`
I cannot get back to Template page by using the back arrow. Well, not always. If I navigate around and try again then it works.

If I go to Partner or Customer the URL is always:
`https://stage.foo.redhat.com:1337/insights/content/repositories`
and I can go back to Templates.

Starle21:
@swadeley 

> "Red Hat repositories" ... /content/repositories?origin=red_hat
I cannot get back to Template page by using the back arrow

Yes, that is expected, it's the same way it worked before. You have to hit go back twice to get back to Templates.

> Partner or Customer the URL is: .../content/repositories

That is also correct. Again the same behavior as before. Only red hat repositories get the origin parameter in the URL.

Note: I updated the color of icons, everything is ready to go :)

swadeley (approved):
ACK

PR #1112, opened 2026-07-20
HMS-9841: Reorder Content nav items & add visual dividers

swadeley:
## Summary
Reorder Content nav items & add visual dividers

HMS-9841 [UI] Remove link to repositories from navigation
## Testing steps
tests pass
Side navbar menu is updated

### DIFF
--- a/deploy/frontend.yaml
+++ b/deploy/frontend.yaml
@@ -40,18 +40,21 @@ objects:
               title: Content
               expandable: true
               routes:
-                - segmentRef:
-                    frontendName: patch
-                    segmentId: patch-advisories-packages
-                - id: repositories
-                  title: Repositories
-                  href: /insights/content/repositories
                 - id: templates
                   title: Templates
                   href: /insights/content/templates
-                - segmentRef:
-                    frontendName: patch
-                    segmentId: patch-systems
+                - id: advisories
+                  title: Advisories
+                  href: /insights/patch/advisories
+                - id: packages
+                  title: Packages
+                  href: /insights/patch/packages
+                - id: repositories
+                  title: Repositories
+                  href: /insights/content/repositories
+                - id: patchSystems
+                  title: Systems
+                  href: /insights/patch/systems
       serviceTiles:
         - section: systemConfiguration
           group: rhel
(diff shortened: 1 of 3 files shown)


xbhouse:
https://issues.redhat.com/browse/HMS-9841

swadeley:
Hi, some have suggested using a line to separate Templates from Repositories:

image

swadeley:
Hi, another example, using Patternfly 6 style (but if you navigate away and back directly to one of the entries below Repositories the line will be gone):

image

swadeley:
Hi

this is what Maria asked for:


image

swadeley:
@sourcery-ai review

swadeley:
@sourcery-ai review

swadeley:
@sourcery-ai review

swadeley:
@sourcery-ai review

xbhouse:
thanks @swadeley :) i'm seeing the dividers but not the reordering in the nav. i just ran `yarn start:stage`, should i run it differently?
Screenshot From 2026-07-29 18-24-54

swadeley:
Hi @xbhouse , 
please try `yarn fec dev --clouddotEnv=stage` 
or just:
` yarn fec dev ` and then select `stage` at the menu prompt,
that should pick up the `deploy/frontend.yaml` changes.

deploy/frontend.yaml
dominikvagner, line 1: what's the reason for not using the segment refs anymore?
the same ordering can be achieved even if we use them 🤔

src/Hooks/useContentNavDivider.ts
dominikvagner, line 1: **this is quite the hack** 😆

_but it sort of works, ~ish?!_, once you visit the repositories or templates pages 🤔 
otherwise you won't get to see these at all (and if you reload, on a different page after visiting ours, then they disappear) 😵‍💫 
which kinda defeats the purpose of the dividers in the first place, as they should make the templates link more visible (to new users), iiuc 🤷🏼  

the correct solution would be to go through "insights chrome"/"feo" or some other platform piece and add support for dividers there 👀 which indeed would be a less straightforward change, but as is, I am kinda against merging this 🚫  
currently this is watching something that we don't control and could change later, and then injecting styling and/or content into it 😬

xbhouse: ah right, the dividers aren't rendered until you actually visit these pages :/ i agree with @dominikvagner that the best place for this would be on the platform side. i'd reach out to that team and see how we can add something like this, i'd think it would likely be in chrome

swadeley: Hi all, OK, I will close this.

Thank you for checking.

PR #1118, opened 2026-07-20
HMS-11054: Add /lightwell/demo with mocked data

jlsherrill:
## Summary

Adds a lightwell demo site that mirrors what we already have but doesn't require a lightwell entitlement 
Search doesn't work, but everything is on one page, so i think thats okay.

## Testing steps
 
launch and go to /lightwell/demo click around.

### DIFF
--- a/src/Pages/Lightwell/mockPackages.ts
+++ b/src/Pages/Lightwell/mockPackages.ts
@@ -1,80 +1,283 @@
-import { RepositoryPackageItem } from 'services/Content/ContentApi';
+import {
+  MavenPackageVersionsListResponse,
+  PythonPackageVersionsResponse,
+  RepositoryPackageItem,
+} from 'services/Content/ContentApi';
 
 const mockPackagesByRepository: Record<string, RepositoryPackageItem[]> = {
+  // Python Remediated
   '33333333-3333-4333-8333-333333333333': [
     {
       name: 'python-requests',
       group: '',
-      versions: ['2.32.3', '2.32.4'],
+      versions: ['2.32.5', '2.32.6'],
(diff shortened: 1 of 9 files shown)


xbhouse:
https://issues.redhat.com/browse/HMS-11054

src/Pages/Lightwell/mockPackages.ts
```diff
@@ -1,80 +1,286 @@
...
+      group: 'org.bouncycastle',
+      versions: ['1.79.0'],
+      latest_releases: [
+        { version: '1.79.0', release: 'rhlw-00001', created_at: '2026-06-05T00:00:00Z' },
+      ],
+    },
+    {
+      name: 'netty-handler',
+      group: 'io.netty',
+      versions: ['4.1.115', '4.1.115', '4.1.114', '4.1.114'],
```
xbhouse: since Ryan's fixing the issue where duplicated versions are returned, maybe we should update the mock data to only include unique versions?

jlsherrill: will do!

xbhouse:
there's one other thing on the releases tab - packages with multiple releases per version aren't appended correctly on the demo site, but Ryan's PR should also fix that. we could update the mock data to only include one release per version or wait to do the prod push until the fix for that is in too?

xbhouse (dismissed):
if the above comment shouldn't block this feel free to merge, everything else looks good :)

jlsherrill:
I think the latest update (adding the release to the version strings in the builds list resolve it

xbhouse (approved):

PR #1135, opened 2026-07-31
HMS-9841: Reorder NavNar menu items

swadeley:
## Summary
HMS-9841:[UI] Re-order links to repositories from navigation

image

## Testing steps
test pass
Menu is as per picture

### DIFF
--- a/deploy/frontend.yaml
+++ b/deploy/frontend.yaml
@@ -40,18 +40,21 @@ objects:
               title: Content
               expandable: true
               routes:
-                - segmentRef:
-                    frontendName: patch
-                    segmentId: patch-advisories-packages
-                - id: repositories
-                  title: Repositories
-                  href: /insights/content/repositories
                 - id: templates
                   title: Templates
                   href: /insights/content/templates
-                - segmentRef:
-                    frontendName: patch
-                    segmentId: patch-systems
+                - id: advisories
+                  title: Advisories
+                  href: /insights/patch/advisories
+                - id: packages
+                  title: Packages
+                  href: /insights/patch/packages
+                - id: repositories
+                  title: Repositories
+                  href: /insights/content/repositories
+                - id: patchSystems
+                  title: Systems
+                  href: /insights/patch/systems
       serviceTiles:
         - section: systemConfiguration
           group: rhel
--- a/deploy/frontend.yaml
+++ b/deploy/frontend.yaml
@@ -43,18 +43,15 @@ objects:
                 - id: templates
                   title: Templates
                   href: /insights/content/templates
-                - id: advisories
-                  title: Advisories
-                  href: /insights/patch/advisories
-                - id: packages
-                  title: Packages
-                  href: /insights/patch/packages
+                - segmentRef:
+                    frontendName: patch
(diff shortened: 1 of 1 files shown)


deploy/frontend.yaml
dominikvagner, line 1: what's the reason for not using the segment refs anymore?
the same ordering can be achieved even if we use them 🤔

xbhouse:
https://issues.redhat.com/browse/HMS-9841

deploy/frontend.yaml
swadeley: Hi @dominikvagner, no reason other than my ignorance of their importance.  They are back now and here is new screen capture:

image

swadeley: Now I see space between the menu entries.

swadeley: and the order does not display correctly locally.
I restarted FE and  am using `yarn fec dev --clouddotEnv=stage`

deploy/frontend.yaml
```diff
@@ -40,15 +40,15 @@ objects:
               title: Content
               expandable: true
               routes:
+                - id: templates
+                  title: Templates
+                  href: /insights/content/templates
                 - segmentRef:
                     frontendName: patch
                     segmentId: patch-advisories-packages
```
Starle21, line 48: The proposed order with this change is:

Templates
Advisories
Packages
Repositories
Systems

But the order that we want to land on is:

Templates
Advisories
Repositories
Packages
Systems

Which means we also need to do changes on the Patch side due to the segmentId: patch-advisories-packages, which groups Advisories and Packages together right now.
So we need to also go to Patch to change it there, since we want to split this group.

Starle21 (changes requested):

deploy/frontend.yaml
Starle21: We also want to show one divider between Templates - Advisories. But we need to ask the platform team if there is support for a divider at all and how to do that.

swadeley: Hi @Starle21 , 

the order of the entries was given to me by Maria and I noted it in the Jira[1]
With this PR we just want to test if the order will build as we want it to in stage (easy to change once we know it will).

Dividers are out of scope for this PR as we already have @dominikvagner 's feedback that we should not override such things but ask Chrome Fronted team to handle that[2]. We need to the order change before they get to work.


[1] https://redhat.atlassian.net/browse/HMS-9841?focusedCommentId=17752643
[2] https://redhat.atlassian.net/browse/RHCLOUD-49804

MariSvirik:
Order: 

Templates
Advisories 
Packages
Repositories
Systems

Relocating only Templates keeps this update minimal and easy for users before the larger platform-wide Insights navigation change.

This structure keeps *Advisories* and *Packages* adjacent as co-equal content types, mirroring both Red Hat Satellite and the Insights System detail view (and current structure). Repositories as a group of packages and advisories, are being given lower priority than objects itself and therefore moved to the bottom. Systems are the last to maintain alignment with the rest of the Insights navigation.

Starle21 (approved):
Approved upon Maria's comments. Good job! 🎉

